PR #22, opened 2015-06-13
pass transaction through sendPushNotification methods

robbiecarlton:
Please take a look. I passed options (which is currently just {transacting: trx}) through everything.


levity:
Looks good, :shipit:

PR #24, opened 2015-06-16
catch errors in PushNotificationController.js

robbiecarlton:


levity:
hmm -- the calls to forbidden shouldn't be in the controller, but in policies. if all you want to verify is that req.session.userId is present, you can just change the entries corresponding to the routes in policies.js to ['sessionAuth']

levity:
also, for the sake of readability and minimizing indentation, i'd prefer that you short-circuit in the beginning of the method, viz:
```js
var token = req.param('deviceToken');
if (!token) {
  return res.serverError('no device token');
}
```

robbiecarlton:
Done and done

PR #27, opened 2015-06-29
Ios logins2

robbiecarlton:


api/controllers/SessionController.js
```diff
@@ -25,8 +25,15 @@ var findCommunity = function(req) {
   });
 };
 
-var finishOAuth = function(service, req, res, next) {
-  passport.authenticate(service, function(err, profile, info) {
+var finishOAuth = function(strategy, req, res, next) {
+  var service = strategy;
+    if (strategy=='facebook-token') {
```
levity: indentation is off for this if-else clause. otherwise, :+1:

levity: and if i really wanted to be a stickler i would say "use `===` instead of `==`"

PR #29, opened 2015-07-06
Disabledevice

robbiecarlton:


api/controllers/PushNotificationController.js
```diff
@@ -23,19 +23,37 @@ module.exports = {
...
 
+  disableDevice: function (req, res) {
+    var token = req.param('deviceToken');
+    if (!token) {
+      return res.serverError('no device token');
+    }
+
+    return Device.forge({
+      token: token,
+      user_id: req.session.userId
```
levity: You can avoid the SELECT entirely here:
`Device.query().where({token: token}).update({enabled: false})`

config/routes.js
```diff
@@ -146,6 +146,7 @@ module.exports.routes = {
   'GET    /noo/logout':                                   'SessionController.destroy',
 
   'POST    /noo/pn/add-device':                           'PushNotificationController.addDevice',
+  'POST    /noo/pn/disable-device':                       'PushNotificationController.disableDevice',
```
levity: The more "RESTful" way of setting this up, following what I've been trying to do with the code so far, would be to have `POST /noo/device => DeviceController.create` and `DELETE /noo/device/:token => DeviceController.destroy`.

robbiecarlton: cool, so would it work if it were 
`POST /noo/device/:token => DeviceController.create` 
and 
`DELETE /noo/device/:token => DeviceController.destroy`
as the create method needs the token as well?

levity: By convention, the id (in this case the token) is only present in the URL
if you're updating an existing record. it doesn't mean you can't still pass
the token in as a query parameter or POST body parameter.

robbiecarlton:
How's this look?

levity:
looks good, just remove the log lines and then feel free to merge

PR #37, opened 2015-09-21
Android push 2

robbiecarlton:


robbiecarlton:
@levity - could you take a look at this when you get the chance, thanks!

migrations/schema.sql
```diff
@@ -16,6 +16,42 @@ SET default_tablespace = '';
 SET default_with_oids = false;
 
 --
+-- Name: active_admin_comments; Type: TABLE; Schema: public; Owner: -; Tablespace: 
+--
+
+CREATE TABLE active_admin_comments (
```
levity: it looks like your local copy of the database is out of sync -- you can either remove the tables listed here individually, until the schema.sql contains only your changes, or just drop your whole dev db and load a new version from production

levity:
apart from the schema.sql, this looks good.

note that after this gets deployed, you'll have to run "heroku run npm run migrate"

migrations/schema.sql
robbiecarlton: If I drop the whole db, load a new one from production, then I should
npm run migrate
to run my migrations and then
 ./migrations/dump_schema.sh > ./migrations/schema.sql
and add the new schema to this pull request
?

levity: yep!

robbiecarlton:
ok, updated schema. looks right to me. Good to merge?

PR #38, opened 2015-09-24
Add missing .env bits

ijt:
Without these, hylo-node crashes.


levity:
:+1:

levity:
Btw, the reason CircleCI builds fail on pull requests is because it doesn't expose environment variables on builds for forks. I guess that's for security reasons, since any forker could push a build that just harvested environment variables.

PR #40, opened 2015-10-07
Email post creation

robbiecarlton:


lib/community/solicitor.js
```diff
@@ -0,0 +1,64 @@
...
+var queueEmail = function (emailData) {
+  return Queue.classMethod('Email', 'sendPostSolicitation', emailData)
+}
+
+Solicitor.sendToCommunity = function (communityId) {
+  return Community.find(communityId, {withRelated: ['users']})
+  .then(community => {
+    sails.log.debug('Solicitor - sending emails to ', community.get('name'))
+    return Promise.map(community.relations.users.models, user =>
+      queueEmail(emailData(user, community)))
```
levity: One thing I would suggest is breaking out the functionality to send a single email for a community to one address into a separate method, so that it's easy to test in production

levity:
Other than my suggestion above, it looks good to merge

levity:
One more thing -- I think you need to handle the setting of the jsonb settings column carefully in CommunityController. Look at the 'preserves existing settings keys' test in User.test.js for what I mean

lib/community/solicitor.js
```diff
@@ -0,0 +1,64 @@
...
+  })
+}
+
+Solicitor.sendWeekly = function () {
+  return Community.query('whereRaw', 'community.settings @> \'{"email_solicitation": true}\'')
```
levity: it's a bit more readable to use `->>` here, i think

PR #41, opened 2015-10-07
Shadow profiles

robbiecarlton:


lib/community/digest.js
```diff
@@ -312,7 +335,7 @@ Digest.test = function (communityId, timeAmount, timeUnit, userId) {
   var now = moment()
   var then = moment().subtract(timeAmount, timeUnit)
   return Community.find(communityId).then(community => {
-    var digest = new Digest(community, then, now)
+    var digest = new Digest(community, then, now, "daily")
```
levity: I'm noticing a code smell here and wanted to reflect on what it means... 

Adding the "daily" param is strange because the test method allows you to parameterize the amount of time you're generating the digest for, so having to specify "daily" or "weekly" doesn't make sense. And this leads me to the observation that the list of recipients really shouldn't be something that the digest class itself generates; it could be something that gets passed in as an option, and we would use simple functions to package up the generating of recipients for the common daily and weekly cases.

Notice also that `sendDaily` and `sendWeekly` have a lot of duplication between them.

In general I've noticed that when I jump right into making a class with Javascript, rather than a collection of functions, I'm more likely to end up with something that feels hard to work with.

Anyway, this isn't so bad that it really cries out to be refactored right away. So feel free to merge if you don't want to refactor.

levity: btw, https://codeclimate.com/github/Hylozoic/hylo-node/lib/community/digest.js is interesting if you haven't seen it yet

robbiecarlton: makes sense. It's a little daunting to read ;)
I'm gonna merge this and then add a task in the backlog to refactor

PR #44, opened 2015-10-08
Email posts 2

robbiecarlton:


lib/community/solicitor.js
```diff
@@ -46,13 +57,17 @@ var queueEmail = function (emailData) {
 }
 
 Solicitor.sendToUser = function (user, community) {
-  return queueEmail(emailData(user, community))
+  if (user.get('settings').email_solicitation_preference) {
+    return queueEmail(emailData(user, community))
+  } else {
+    return Promise.resolve()
+  }
```
levity: Another way to do this is to wrap the entire method in `Promise.method` and then just return nothing if the user doesn't have the setting turned on

robbiecarlton: Do you have a preference for the Promise.method way?
The above is easy for me to read

robbiecarlton:
@levity good to merge?

lib/community/solicitor.js
levity: It's fine, just FYI. Would be a bit more readable to use Promise.method in cases with more branching.

levity:
:shipit:

PR #45, opened 2015-10-13
add post title to push notifications

robbiecarlton:


robbiecarlton:
@levity good?

levity:
I don't think we should be sending notifications about your own welcome posts.

levity:
Sending them about other people's welcome posts is also... hmm, maybe something that should have its own setting. Maybe not a big deal at the moment, and actually probably useful for early members / moderators in a community, but potentially quite noisy.

levity:
@edwardwest what do you think?

robbiecarlton:
Ok, it no longer sends notifications on welcome posts. Let me know if you want it to send them on other peoples welcome posts, otherwise I'll merge...

levity:
:shipit:

PR #47, opened 2015-10-22
Email Solicitation Prefs default to true in user and communtiy

robbiecarlton:


robbiecarlton:
and then when this has been merged, I'll run this in the console on the production server:

https://gist.github.com/robbiecarlton/628d98e32f4f13c27a8a#file-updatecommunities-js

levity:
instead of calling it 'email_solicitation_preference', can we just name it like we would any other boolean variable, e.g. 'receives_email_prompts'?

robbiecarlton:
sure. I modeled the name after send_email_preference in the user model. Not totally clear what "name it like any other boolean variable" means, but I can rename it to 'receives_email_prompts'

levity:
yeah, that other naming choice predates me. i just mean that i've tried to
name boolean attributes, and functions that return booleans, with a verb
and an object describing a state, e.g. isCodeValid, isCreator, sometimes
dropping the verb if the phrase clearly indicates a state by itself.
something ending in 'preference' would be more appropriate to a 3+-valued
variable.

robbiecarlton:
Ok, renamed it all from solicitation to prompts
see also: https://github.com/Hylozoic/hylo-frontend/pull/73
which should be merged at the same time

levity:
:+1:

PR #49, opened 2015-10-23
RSVP for events

robbiecarlton:


robbiecarlton:
@levity Schema.sql is a little off, but only by the order of two fields. Does that matter?
If not, ready to merge

api/models/User.js
```diff
@@ -56,6 +56,10 @@ module.exports = bookshelf.Model.extend({
     return this.hasMany(Post)
   },
 
+  eventsResponsedTo: function () {
```
levity: this should be eventsRespondedTo

api/controllers/PostController.js
```diff
@@ -348,6 +348,34 @@ module.exports = {
...
+
+    EventResponse.query(qb => {
+      qb.where({user_id: userId, post_id: post.id})
+      qb.orderBy('created_at', 'desc')
+      return qb
+    }).fetchAll()
+    .then(eventResponses => {
+      if (eventResponses.length > 0) {
+        var lastResponse = eventResponses.models[0].get('response')
+        return Promise.map(eventResponses.models, eventResponse => { eventResponse.destroy() })
```
levity: no need for curly braces around eventResponse.destroy() -- in fact, since it's a one-liner, i would usually name the argument something briefer like `r` or `resp`

levity: also, if you're going to destroy and re-create, why not update instead?

api/controllers/PostController.js
```diff
@@ -348,6 +348,34 @@ module.exports = {
...
+        .then(() => {
+          if (response !== lastResponse) {
+            return EventResponse.create(post.id, {responderId: userId, response: response})
+          } else {
+            return Promise.resolve()
```
levity: you don't need to use Promise.resolve() for the else case here, since you're already in a function that's being passed to `then` -- just do nothing

api/controllers/PostController.js
```diff
@@ -348,6 +348,34 @@ module.exports = {
...
+
+  respond: function (req, res) {
+    var userId = req.session.userId
+    var post = res.locals.post
+    var response = req.param('response')
+
+    EventResponse.query(qb => {
+      qb.where({user_id: userId, post_id: post.id})
+      qb.orderBy('created_at', 'desc')
+      return qb
```
levity: you don't need to return qb -- Bookshelf's query method works entirely through side-effects on the argument

levity:
the minor quirk in schema.sql doesn't matter -- it does mean our two dev databases are out of sync, but this should be resolved after we've both dumped production copies into our development environments

robbiecarlton:
fixed. much simpler now :)
good to merge?

robbiecarlton:
@levity ok to merge?

levity:
:+1:

PR #55, opened 2015-11-12
Add event times to events in digest

robbiecarlton:


robbiecarlton:
screen shot 2015-11-11 at 5 57 49 pm

levity:
reduce the vertical gap between time and location and we're probably good
-- maybe (mayyyybe) make the font size the same as the footer line?

levity:
also pls confirm that we don't get a gap in the template for posts that
don't have time and location

robbiecarlton:
Yes no gap when there's no time or location. I narrowed the vertical gap and kept the font size the same for now. Gonna merge. If we want to make more style changes, it's all in the template.

screen shot 2015-11-12 at 12 22 35 pm

PR #57, opened 2015-11-18
(WIP) Projects use media table for images and videos

robbiecarlton:


robbiecarlton:
@levity, lmk and I'll merge and update the production DB

api/controllers/ProjectController.js
```diff
@@ -1,188 +1,230 @@
...
+  'media'
+]
+
+var projectAttributes = function (project) {
+  var attrs = project.toJSON()
+  attrs = _.extend(attrs, {
+    contributor_count: project.relations.contributors.length,
+    open_request_count: project.relations.posts.length
+  })
+  attrs = _.extend(attrs, mediaAttributes(project))
```
levity: why extend twice? also, _.extend operates in-place, so you don't need to assign

levity: with extend you can add as many arguments as you want and they'll all get merged into the first one

api/controllers/ProjectController.js
```diff
@@ -1,188 +1,230 @@
...
+  return attrs
+}
+
+var mediaAttributes = function (project) {
+  var attrs = project.toJSON()
```
levity: better not to have project.toJSON involved here, because it's redundant when this is used in projectAttributes. and then in findOne, just continue to use project.toJSON and extend it with mediaAttributes

api/controllers/ProjectController.js
```diff
@@ -1,188 +1,230 @@
...
+    } else if (req.param('unpublish')) {
+      updatedAttrs.published_at = null
+    }
+
+    if (_.has(updatedAttrs, 'title')) {
+      updatedAttrs.slug = makeSlug(updatedAttrs.title)
+    }
+
+    createMedia(project, mediaAttrs)
+      .then(() => {
```
levity: don't indent `.then` please

levity:
want to add some tests?

robbiecarlton:
Sure, I'll add some tests

robbiecarlton:
I had to keep project.toJSON in mediaAttributes, because I call it in two different places. In one place it's merged with projectAttributes, but in the other it's not, so it needs it's own set of the projects attributes. 
I can think of a couple of ways of simplifying, but I'd like to hear your thoughts...

levity:
I'm suggesting that in the other place (`findOne`), you keep using `project.toJSON` like it was before and just add mediaAttributes as an additional argument to the `merge` that's already there.

robbiecarlton:
got it

robbiecarlton:
@levity good to merge?

levity:
:cool: :shipit:

Issue #50, opened 2015-10-27
Ability to notify via slack may lead to significant increase in interaction


Connoropolous:
My perspective on hylo is that, unless you have the app installed, the 'pace of conversation' is slower than ideal. Getting it into a dimension of closer to realtime interactivity, almost chat, could really increase engagement. A great starting point for this, I think, would be 'push to slack' notifications of activity. I've built these kinds of things before, what this requires is
- [ ] to decide upon what models from hylo can be tracked in slack: a post, a community, a network, a project? 
- [ ] the addition of minimal UI to accept the depositing of a 'webhook' url, by a user
- [ ] documentation somewhere for users of what to do to retrieve a slack webhook url
- [ ] extension of the current notifications code to post to slack webhooks, when present on a resource

Once again, I may be interested to actually write and submit a PR for this, but I would want to get feedback here first.

Connoropolous:
Slack is where the people are, and where their attention is.

edwardwest:
Yes-- agreed-- to a certain extent having the mobile app installed helps
with this, as there are push notifications for every post-- but there is
more that could be done.

I would love it if you took a look at how to do this, Connor--

>
>    - to decide upon what models from hylo can be tracked in slack: a
>    post, a community, a network, a project?
>
>
I think for the first rollout, the best would be to track the "Community"
in a slack group-- this will catch all of the Projects as well, but not be
potentially overwhelming with Network activity


>
>    - the addition of minimal UI to accept the depositing of a 'webhook'
>    url, by a user
>    - documentation somewhere for users of what to do to retrieve a slack
>    webhook url
>    - extension of the current notifications code to post to slack
>    webhooks, when present on a resource
>
>
(I'll let Lawrence or Robbie weigh in on these three)

>
>
> Once again, I may be interested to actually write and submit a PR for
> this, but I would want to get feedback here first.
>
Thanks for taking a look at this!
Edward

Connoropolous:
Progress underway!

<img width="766" alt="screen shot 2015-11-19 at 1 02 06 pm" src="https://cloud.githubusercontent.com/assets/1409121/11279331/dcfb9a2c-8ebd-11e5-8b9d-daf6a620a25d.png">

<img width="697" alt="screen shot 2015-11-19 at 1 02 18 pm" src="https://cloud.githubusercontent.com/assets/1409121/11279337/e84d1df6-8ebd-11e5-9d21-f4e2f0f031fb.png">

- [x] migration written
[...]

PR #65, opened 2015-11-19
Send notifications of new posts in a community to slack

Connoropolous:
Woohoo! This was fun :)

I'm sure you'll have a few little pointers about things that need to be resolved before it gets merged, let me know what they are!

relates to front end PR https://github.com/Hylozoic/hylo-frontend/pull/85 
and issue #50 

- [x] point to a better image asset

screen shot 2015-11-19 at 5 23 12 pm

### DIFF
--- a/api/models/Post.js
+++ b/api/models/Post.js
@@ -240,18 +240,24 @@ module.exports = bookshelf.Model.extend({
         var users = uniqById(_.flatten(usersWithDupes))
 
         _.remove(users, user => user.get('id') === creator.get('id'))
-        return Promise.map(users, (user) => {
-          if (!user.get('push_new_post_preference')) return
-          if (post.isWelcome()) return
-          var userCommunities = user.relations.communities.models
-          var postCommunitiesIds = communities.models.map(community => community.get('id'))
-          var community, path, alertText
-          community = _.find(userCommunities, community => _.contains(postCommunitiesIds, community.get('id')))
-          if (!community) return
-          path = url.parse(Frontend.Route.post(post, community)).path
-          alertText = PushNotification.textForNewPost(post, community, user.get('id'))
-          return user.sendPushNotification(alertText, path)
-        })
+        return Promise.join(
+          Promise.map(communities.models, (community) => {
+            if (!community.get('slack_hook')) return
+            return Community.sendSlackNotification(community.id, post); 
+          }),
+          Promise.map(users, (user) => {
+	    if (!user.get('push_new_post_preference')) return
+	    if (post.isWelcome()) return
+	    var userCommunities = user.relations.communities.models
+	    var postCommunitiesIds = communities.models.map(community => community.get('id'))
+	    var community, path, alertText
+	    community = _.find(userCommunities, community => _.contains(postCommunitiesIds, community.get('id')))
+	    if (!community) return
+	    path = url.parse(Frontend.Route.post(post, community)).path
+	    alertText = PushNotification.textForNewPost(post, community, user.get('id'))
+	    return user.sendPushNotification(alertText, path)
+	  })
+        )
       })
   },
 
--- a/api/models/Post.js
+++ b/api/models/Post.js
@@ -241,22 +241,22 @@ module.exports = bookshelf.Model.extend({
 
         _.remove(users, user => user.get('id') === creator.get('id'))
         return Promise.join(
-          Promise.map(communities.models, (community) => {
+          Promise.map(communities.models, community => {
             if (!community.get('slack_hook')) return
-            return Community.sendSlackNotification(community.id, post); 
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -78,6 +78,7 @@
     "randomstring": "^1.1.0",
     "rc": "~0.5.0",
     "request": "^2.51.0",
+    "request-promise": "^1.0.2",
```
levity: interesting -- i've just been using bluebird's `promisify` on request.get and request.post, so i'm not sure if we need to add this dependency

Connoropolous: Ok cook. Yeah, that should work. Can you point me to example? I'm pretty new to promises fyi

migrations/20151119113256_community_slackhook.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table('community', function(table) {
+    table.text('slack_hook');
```
levity: interesting note -- i was going to suggest that `string` is more appropriate here than `text`, but then i looked it up and in Postgres [there's actually no difference](http://www.postgresql.org/docs/8.4/static/datatype-character.html). so carry on :)

Connoropolous:
We should point to a better image asset than the one I pointed to.

api/models/Post.js
```diff
@@ -236,18 +236,24 @@ module.exports = bookshelf.Model.extend({
...
-          alertText = PushNotification.textForNewPost(post, community, user.get('id'))
-          return user.sendPushNotification(alertText, path)
-        })
+        return Promise.join(
+          Promise.map(communities.models, (community) => {
+            if (!community.get('slack_hook')) return
+            return Community.sendSlackNotification(community.id, post); 
+          }),
+          Promise.map(users, (user) => {
+	    if (!user.get('push_new_post_preference')) return
```
levity: your indentation seems to have gotten a little wonky here

package.json
levity: Basically you just do:

```js
var Promise = require('bluebird')
var request = require('request')
var get = Promise.promisify(request.get)

get(url).spread((resp, body) => ...).catch(err => ...)
```

FYI: in bluebird 3 (which I started using in a different project; here we're still on v2) you have to do `promisify(get, {multiArgs: true})` in order to get both `resp` and `body` into the `then` handler

api/models/Post.js
```diff
@@ -236,18 +236,24 @@ module.exports = bookshelf.Model.extend({
...
-          var postCommunitiesIds = communities.models.map(community => community.get('id'))
-          var community, path, alertText
-          community = _.find(userCommunities, community => _.contains(postCommunitiesIds, community.get('id')))
-          if (!community) return
-          path = url.parse(Frontend.Route.post(post, community)).path
-          alertText = PushNotification.textForNewPost(post, community, user.get('id'))
-          return user.sendPushNotification(alertText, path)
-        })
+        return Promise.join(
+          Promise.map(communities.models, (community) => {
```
levity: just a little style note: you don't need to have `( )` for single-argument methods

levity:
re: images: you could try these two

http://cdn.hylo.com/misc/hylo-logo-teal-on-transparent.png
http://cdn.hylo.com/misc/hylo-logo-white-on-teal-circle.png

Connoropolous:
@levity I'm going to come back and finish this soon! oops! took me a bit. 

@edwardwest , the newly created CTA slack + CTA hylo is another example of a community who could benefit from this integration

Connoropolous:
I am going to try to catch this up to master and then finish it :) finally. might ping for help if I get stuck doing migrations or something

levity:
:cool:

Connoropolous:
using the default image, it looks like this now

screen shot 2016-02-10 at 11 39 19 am

Connoropolous:
This is ready for another review now :) 

The only steps needed to make this work would be 
- [ ] running the migration
- [ ] setting the HYLO_SLACK_ICON env variable

api/services/Slack.js
```diff
@@ -0,0 +1,34 @@
+var format = require('util').format
+var Promise = require('bluebird')
+var request = require('request')
+var post = Promise.promisify(request.post)
```
harlantwood: You may like to consider using the `request-promisify` module, which combines request + bluebird.

Connoropolous: haha, actually I had ben doing it that way, but Lawrence mentioned he'd been doing it this way in other places in the app, so I figured I'd just keep on that way :) cheers though Harlan. Keen for other pointers whenever you have them

harlantwood: Ah cool.  Consistency is worth a lot : )

PR #78, opened 2016-02-09
Update user_id if device already exists

robbiecarlton:


test/unit/controllers/DeviceController.test.js
```diff
@@ -47,5 +49,35 @@ describe('DeviceController', () => {
...
+
+      return Device.fetchAll()
+      .then(devices => Promise.map(devices.models, device => device.destroy()))
+      .then(() => Device.forge({
+        token: req.param('token'),
+        platform: platform,
+        version: version,
+        user_id: fixtures.u2.id
+      })
+      .save()
```
levity: nitpick: this should be on the line above, right after `})`. when it's on its own line the indentation suggests its outside of the `.then` above it, but it's actually still nested inside it.

likewise, i think you want to add another `)` after `.save()` and remove it from line 80 so that the indentation below makes sense

levity:
apart from my comment, :+1:

test/unit/controllers/DeviceController.test.js
robbiecarlton: thanks! I knew there was something off but I couldn't make it right.

PR #80, opened 2016-02-10
Network Creation

robbiecarlton:


robbiecarlton:
@levity ok, can I merge this one?

api/controllers/SearchController.js
```diff
@@ -1,11 +1,21 @@
...
+    .then(cs => Promise.map(cs.models, c => c.id))
   } else {
     return Promise.join(
       Network.activeCommunityIds(req.session.userId),
       Membership.activeCommunityIds(req.session.userId)
-    ).then(ids => _(ids).flatten().uniq().value())
+    ).then(dupIds => {
+      var ids = _(dupIds).flatten().uniq().value()
+      if (req.param('moderated')) {
+        return Promise.filter(ids, id => Membership.hasModeratorRole(req.session.userId, id))
```
levity: Promise.filter... neat

api/controllers/SearchController.js
```diff
@@ -1,11 +1,21 @@
 var findCommunityIds = Promise.method(req => {
   if (req.param('communityId')) {
     return [req.param('communityId')]
+  } else if (req.param('type') === 'communities' && req.param('moderated') && Admin.isSignedIn(req)) {
+    return Community.fetchAll()
+    .then(cs => Promise.map(cs.models, c => c.id))
```
levity: this can just be `.then(cs => cs.pluck('id'))`

api/controllers/SearchController.js
levity: however, this is going to be a pretty slow operation for Ed, because it's one query per membership. what i would suggest instead is to create a new class method on Membership (or add a parameter to `activeCommunityIds`) that only returns the memberships with moderator roles.

api/controllers/SearchController.js
```diff
@@ -92,6 +106,7 @@ module.exports = {
     })).fetchAll({columns: columns}))
     .then(rows => {
       var present
+      var filteredRows = rows
```
levity: i'm guessing this is a vestigial change from a different approach?

api/controllers/NetworkController.js
```diff
@@ -1,7 +1,62 @@
...
+
+  create: function (req, res) {
+    var attrs = _.pick(req.allParams(),
+      'name', 'description', 'slug', 'banner_url', 'avatar_url')
+
+    var network = new Network(_.merge(attrs, {
+      created_at: new Date()
+    }))
+
+    return network.save()
```
levity: wrap all the saves in a transaction

api/controllers/NetworkController.js
```diff
@@ -1,7 +1,62 @@
...
+        }
+        return
+      })
+    ))
+    .then(res.ok)
+    .catch(res.serverError)
+  },
+
+  validate: function (req, res) {
+    var allowedColumns = ['name', 'slug']
```
levity: this code looks familiar :) want to pull it out of here and CommunityController into a new file in services? there's already a good set of tests for it

api/controllers/SearchController.js
robbiecarlton: that's never going to run for Ed though, because he'll be logged in as Admin, so it runs through the
`req.param('type')` === 'communities' && req.param('moderated') && Admin.isSignedIn(req)
branch

Still want me to write the new method?

levity: Ah, right. Yeah, though, might as well take care of this now because it's eventually going to be accessible to all users.

api/services/Validation.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
```
robbiecarlton: @levity something like this?

levity: Don't couple it to req and res though.

robbiecarlton:
@levity - ready to merge?

levity:
:cool:

PR #81, opened 2016-02-10
Use a slack app instead of manual slack configuration

Connoropolous:
Note, I was just trying this to see whether I could do it, but it does seem like it improves the user experience a fair bit, so we may want to hold off on merging the simpler variation if y'all really dig this :)

- [x] catch error cases
- [x] store more than just the webhook url, store the team name, and the url they could go to disable it, for display on the UI
- [x] finish the front end

### DIFF
--- a/api/controllers/CommunityController.js
+++ b/api/controllers/CommunityController.js
@@ -1,8 +1,8 @@
-var Promise = require('bluebird'),
-  request = require('request'),
-  format = require('util').format,
-  get = Promise.promisify(request.get),
-  slackAuthAccess = 'https://slack.com/api/oauth.access';
+var Promise = require('bluebird')
+var request = require('request')
+var format = require('util').format
+var get = Promise.promisify(request.get)
+var slackAuthAccess = 'https://slack.com/api/oauth.access'
 
 module.exports = {
   find: function (req, res) {
@@ -35,7 +35,7 @@ module.exports = {
     Community.find(req.param('communityId'), {withRelated: ['leader']})
     .tap(community => leader = community.relations.leader)
     .then(community => _.merge(community.pick(
-      'welcome_message', 'beta_access_code', 'slack_hook', 'slack_team', 'slack_configure', 'settings'
+      'welcome_message', 'beta_access_code', 'slack_hook_url', 'slack_team', 'slack_configure_url', 'settings'
     ), {
       leader: leader ? leader.pick('id', 'name', 'avatar_url') : null
     }))
@@ -46,7 +46,7 @@ module.exports = {
   update: function (req, res) {
     var whitelist = [
       'banner_url', 'avatar_url', 'name', 'description', 'settings',
-      'welcome_message', 'leader_id', 'beta_access_code', 'location', 'slack_hook', 'slack_team', 'slack_configure'
+      'welcome_message', 'leader_id', 'beta_access_code', 'location', 'slack_hook_url', 'slack_team', 'slack_configure_url'
     ]
     var attributes = _.pick(req.allParams(), whitelist)
     var saneAttrs = _.clone(attributes)
@@ -74,11 +74,10 @@ module.exports = {
     get(url).spread((resp, body) => {
       var parsed = JSON.parse(body);
       Community.find(req.param('communityId')).then(function (community) {
-        var communityToUpdate = new Community({id: req.param('communityId')})
-        communityToUpdate.save({
-          slack_hook: parsed.incoming_webhook.url,
+        community.save({
+          slack_hook_url: parsed.incoming_webhook.url,
           slack_team: parsed.team_name,
-          slack_configure: parsed.incoming_webhook.configuration_url
+          slack_configure_url: parsed.incoming_webhook.configuration_url
         }, {patch: true})
         .then(() => res.redirect(Frontend.Route.community(community) + '/settings?slack=1'))
         .catch(() => res.redirect(Frontend.Route.community(community) + '/settings?slack=0'))
--- a/api/controllers/CommunityController.js
+++ b/api/controllers/CommunityController.js
@@ -29,7 +29,7 @@ module.exports = {
     Community.find(req.param('communityId'), {withRelated: ['leader']})
     .tap(community => leader = community.relations.leader)
     .then(community => _.merge(community.pick(
-      'welcome_message', 'beta_access_code', 'settings'
+      'welcome_message', 'beta_access_code', 'slack_hook', 'settings'
     ), {
       leader: leader ? leader.pick('id', 'name', 'avatar_url') : null
     }))
@@ -40,7 +40,7 @@ module.exports = {
   update: function (req, res) {
     var whitelist = [
       'banner_url', 'avatar_url', 'name', 'description', 'settings',
-      'welcome_message', 'leader_id', 'beta_access_code', 'location'
+      'welcome_message', 'leader_id', 'beta_access_code', 'location', 'slack_hook'
     ]
     var attributes = _.pick(req.allParams(), whitelist)
     var saneAttrs = _.clone(attributes)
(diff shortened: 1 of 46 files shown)


Connoropolous:
step 1
screen shot 2016-02-10 at 3 19 14 pm

step 2
screen shot 2016-02-10 at 12 29 59 pm

step 3
screen shot 2016-02-10 at 12 36 00 pm

step 4
screen shot 2016-02-10 at 2 57 33 pm

Connoropolous:
step 1
screen shot 2016-02-10 at 9 52 12 pm

step 2
screen shot 2016-02-10 at 12 29 59 pm

step 3
screen shot 2016-02-10 at 12 36 00 pm

step 4
screen shot 2016-02-10 at 10 13 11 pm

step 4 (alt)
screen shot 2016-02-10 at 10 19 01 pm


step 5
screen shot 2016-02-10 at 2 57 33 pm

step 6
screen shot 2016-02-10 at 9 54 28 pm

step 7
screen shot 2016-02-10 at 9 51 46 pm

back to step 1

Connoropolous:
This could be good for Hylo as well because with a little bit of work you can register the app in the slack app directory :) https://slack.com/apps

levity:
cool! i'll go over the code and make some suggestions.

migrations/20160210212844_community_slack.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table('community', function(table) {
+    table.text('slack_hook');
+    table.text('slack_team');
+    table.text('slack_configure');
```
levity: what do you think about changing this and the `slack_hook` columns to end in `_url` so it's clearer what's to be stored there?

config/routes.js
```diff
@@ -66,6 +66,7 @@ module.exports.routes = {
   'GET    /noo/community/:communityId':                   'CommunityController.findOne',
   'POST   /noo/community/:communityId':                   'CommunityController.update',
   'GET    /noo/community/:communityId/settings':          'CommunityController.findSettings',
+  'GET    /noo/community/:communityId/settings/slack':    'CommunityController.addSlack',
```
levity: This should probably be a POST, no?

api/services/Slack.js
```diff
@@ -0,0 +1,32 @@
...
+var request = require('request')
+var post = Promise.promisify(request.post)
+
+module.exports = {
+  textForNewPost: function (post, community) {
+    var relatedUser, creator = post.relations.creator
+
+    if (post.isWelcome()) {
+      relatedUser = post.relations.relatedUsers.first()
+      return format('<%s|%s> joined <%s|%s>',
```
levity: by the way, now that we've switched to using Node 5, we can use template strings instead of `format` :)

api/controllers/CommunityController.js
```diff
@@ -1,3 +1,9 @@
+var Promise = require('bluebird'),
+  request = require('request'),
```
levity: please make sure that this matches Javascript Standard Style (no semicolons, each var declaration on its own line, etc)

api/controllers/CommunityController.js
```diff
@@ -55,6 +61,33 @@ module.exports = {
...
+                       process.env.SLACK_APP_CLIENT_ID,
+                       process.env.SLACK_APP_CLIENT_SECRET,
+                       code,
+                       redirect_uri);
+
+    get(url).spread((resp, body) => {
+      var parsed = JSON.parse(body);
+      Community.find(req.param('communityId')).then(function (community) {
+        var communityToUpdate = new Community({id: req.param('communityId')})
+        communityToUpdate.save({
```
levity: you can call `save` on the instance of `community` you get from `find`; no need to do `new Community()` here

migrations/20160210212844_community_slack.js
Connoropolous: for sure

config/routes.js
Connoropolous: I think I was trying it with a post, wasn't working, switched it to get and then changed something else and then it was working. but I think the thing that fixed it wasn't switching from Post to Get, so we could switch it back. Not much need though, this way is equally supported https://api.slack.com/web#basics

api/services/Slack.js
Connoropolous: Nice! Right

api/controllers/CommunityController.js
Connoropolous: Cool. Ok, I will have to look into that and aim for that in future. I'll add another commit changing it

api/controllers/CommunityController.js
Connoropolous: nice, ok

api/services/Slack.js
Connoropolous: Actually, in this case, I like it with format better because it makes the full message clearer right away when you read it, versus having to parse out all the higher complexity variable names inline

Connoropolous:
@levity any thoughts on getting this in, in terms of timing or further work? Looking for some guidance and some insight into 'how Hylo does things' :)

config/routes.js
levity: My main concern is that a GET with no parameters is often a good candidate for HTTP caching, either by the browser or by any sort of proxy in between (at ISP's, etc). So using a POST is the right thing to do because it signals never to use a cache.

levity:
After you fix the GET/POST thing and merge master into this to fix conflicts (let me know if you encounter any confusion while merging) I should be able to merge this one in.

Connoropolous:
Ok merged master in, changed get to post, and tested it all. looking good. :) just need to get that one link in the front end being replaced with the CLIENT_ID

Connoropolous:
To set up an application, and get a client id and secret, just go to: https://api.slack.com/applications/new

It will ask for a picture of 512x512, I made this one, from some image you had linked, but maybe get Damian to make one if this one doesn't suit. 

hylo512

Eventually, after some use testing, we can work on adding a promo page to the site, and doing a few other little things listed here (https://api.slack.com/docs/slack-apps-checklist) if we want to get it live into the app store.

harlantwood:
@Connoropolous this is awesome!  I happen to be building a slack app (in ruby) for Swarmbot at the moment, would be fun to compare note.

In case you missed it, there is an oauth test failing on CI BTW: https://circleci.com/gh/Hylozoic/hylo-node/975

levity:
Oh, don't worry about the failing tests on pull requests. The OAuth tests
require valid keys to be in environment variables but those don't get
shared with forks for security reasons.

harlantwood:
Ah makes sense.  I believe you can turn disable automated builds for PRs if desired.

levity:
Re the GET/POST: I meant we should change it for the endpoint that the hylo-node API exposes. Thanks for the explanation about setting up Slack apps.

Connoropolous:
Ohh, I definitely misunderstood. Anyways, that one can't be a POST because Slack is just doing a redirect to that address, which definitely wouldn't be a POST.

levity:
Oh, I see, I misunderstood too :)

api/controllers/CommunityController.js
```diff
@@ -55,6 +61,35 @@ module.exports = {
...
+        community.save({
+          slack_hook_url: parsed.incoming_webhook.url,
+          slack_team: parsed.team_name,
+          slack_configure_url: parsed.incoming_webhook.configuration_url
+        }, {patch: true})
+        .then(() => res.redirect(Frontend.Route.community(community) + '/settings?slack=1'))
+        .catch(() => res.redirect(Frontend.Route.community(community) + '/settings?slack=0'))
+      })
+    }).catch(err => {
+      res.redirect(Frontend.Route.community(community) + '/settings?slack=0')
```
robbiecarlton: `community` is out of scope and so not defined in the catch handler
also, Standard JS says if you declare an err you have to handle it, so should be
`catch(() => {`
instead

PR #87, opened 2016-02-23
Add auto-update-info route

robbiecarlton:
@levity Adding a route for the mobile apps to get info about whether they should force update.
Figured I'd make a new controller for it. Let me know if you want this somewhere else.

### DIFF
--- a/config/routes.js
+++ b/config/routes.js
@@ -165,5 +165,7 @@ module.exports.routes = {
 
   'POST    /noo/subscription':                            'SubscriptionController.create',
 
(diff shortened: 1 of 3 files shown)


config/routes.js
```diff
@@ -165,5 +165,7 @@ module.exports.routes = {
 
   'POST    /noo/subscription':                            'SubscriptionController.create',
 
+  'GET     /noo/auto-update-info':                        'MobileAppController.updateInfo',
```
levity: maybe prefix this with `mobile`?

PR #88, opened 2016-02-23
add active column to community table

robbiecarlton:


robbiecarlton:
@levity I want to do this in a separate PR so that I can set all current communities to active before adding any filtering code...

PR #91, opened 2016-02-25
Add live-status route with new_notification_count

robbiecarlton:


robbiecarlton:
@levity can I merge?

api/controllers/LiveStatusController.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+
+  liveStatus: function (req, res) {
```
levity: this can just be `show`, which follows the convention of controller methods being verbs that operate on the noun that the controller is named after -- `show` for getting a single item with no side effects, `index` for showing multiple items, `create` and `update` as you would expect

levity:
other than that, :+1:

PR #93, opened 2016-03-01
Live post notifications 2

robbiecarlton:


api/controllers/PostController.js
```diff
@@ -392,6 +400,36 @@ module.exports = {
       }
     })
     .then(() => res.ok({}), res.serverError)
-  }
+  },
+
+  checkFreshnessForCommunity: function (req, res) {
+    var foundPosts
+    var resolve = value => foundPosts = value
```
robbiecarlton: this smells to me. I feel like there's a way to do it that's more Promisey, but I can't quite work it out. Whatever it is would likely require another refactor for findPosts. Any thoughts?

robbiecarlton:
@levity will you take a look when you get a minute, and see what you think of this general approach?
Everything else would be a duplication of what I've done in *forCommunity

api/controllers/PostController.js
levity: the way that seems most straightforward to me is that you send as parameters the list of post ids you have along with the updated_at timestamp for each. then all you need to do on the server side is `Post.query().where('id', 'in', thoseIds).filter(p => p.get('updated_at') > originalUpdatedAts[p.id])` (pseudocode, but i hope you get the idea) and return `true` if that list is non-empty.

then you avoid using `findPosts` at all, because you don't need to build the original query from scratch, and you don't need any of the presentation logic.

robbiecarlton: but i need to call findPosts with the specific query for the case where
there is a NEW post id. Which is likely the most common case.

On Mon, Feb 29, 2016 at 4:26 PM, Lawrence Wang <notifications@github.com>
wrote:

> In api/controllers/PostController.js
> <https://github.com/Hylozoic/hylo-node/pull/93#discussion_r54504293>:
>
> > @@ -392,6 +400,36 @@ module.exports = {
> >        }
> >      })
> >      .then(() => res.ok({}), res.serverError)
> > -  }
> > +  },
> > +
> > +  checkFreshnessForCommunity: function (req, res) {
> > +    var foundPosts
> > +    var resolve = value => foundPosts = value
>
> the way that seems most straightforward to me is that you send as
> parameters the list of post ids you have along with the updated_at
> timestamp for each. then all you need to do on the server side is Post.query().where('id',
> 'in', thoseIds).filter(p => p.get('updated_at') > originalUpdatedAts[p.id
> ]) (pseudocode, but i hope you get the idea) and return true if that list
> is non-empty.
>
> then you avoid using findPosts at all, because you don't need to build
> the original query from scratch, and you don't need any of the presentation
> logic.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Hylozoic/hylo-node/pull/93/files#r54504293>.
>



-- 
Robbie Carlton

719.358.3482
<robbie@authenticmanprogram.com>

levity: oh, right. oops. well, then you'll want to refactor `findPosts` to separate the query logic from the presentation logic, and use only the former for these routes.

robbiecarlton: right, makes sense

robbiecarlton:
@levity something a bit weird is happening here.
the code itself works exactly as intended when used by the redux frontend.
each of the new tests (checkFreshnessFor* tests) works by itself, with no warnings.
when I run all of the tests together, they all pass, however I get warnings suggesting that the assertions are not actually being called.
If you get a chance to checkout it out and see what happens when you run the tests, that would be great.

I'm gonna give it a break and switch to redux for a bit.

levity:
it's because you're waiting for a promise but you're not returning that promise in the test function. so, end the test function with `return PostController.checkFreshness...` and make sure the controller function also returns any promises it sets up.

robbiecarlton:
thanks! that fixed it.

On Wed, Mar 2, 2016 at 5:01 PM, Lawrence Wang <notifications@github.com>
wrote:

> it's because you're waiting for a promise but you're not returning that
> promise in the test function. so, end the test function with return
> PostController.checkFreshness... and make sure the controller function
> also returns any promises it sets up.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Hylozoic/hylo-node/pull/93#issuecomment-191520624>.
>



-- 
Robbie Carlton

719.358.3482
<robbie@authenticmanprogram.com>

robbiecarlton:
@levity - good to merge?

api/controllers/PostController.js
```diff
@@ -392,6 +421,19 @@ module.exports = {
       }
     })
     .then(() => res.ok({}), res.serverError)
-  }
+  },
 
+  checkFreshnessForCommunity: createCheckFreshnessAction(queryForCommunity, 'posts'),
```
levity: can you iterate through the different types here and programmatically generate both the `find` and `freshness` methods? it would be way more concise

robbiecarlton:
Done it two different ways, both iterating, and just generating the method names by hand, not sure which I prefer.
Check out this commit for the other version: https://github.com/Hylozoic/hylo-node/commit/510125c045af8b91bcb1a6f6b63ad2d2524c8b08

robbiecarlton:
@levity - like this?

levity:
yay! so much cleaner

PR #95, opened 2016-03-10
Allow Admin to modify any communities network membership

robbiecarlton:


api/controllers/NetworkController.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
       .tap(network => Promise.map(req.param('communities'), communityId =>
         Membership.hasModeratorRole(req.session.userId, communityId)
```
levity: this line is missing a `return`

api/controllers/NetworkController.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
...
         .then(isModerator => {
-          if (isModerator) {
+          if (isModerator || Admin.isSignedIn(req)) {
             return Community.find(communityId)
```
levity: this can be done without the community lookup -- just `Community.query().where('id', communityId).update({network_id: network.id})`

api/controllers/NetworkController.js
robbiecarlton: it's a oneline function literal, just broken into 2 lines. So it has implicit return.

api/controllers/NetworkController.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
     var setNetworkIdIfModerator = (communityId, networkId, trx) => {
       return Membership.hasModeratorRole(req.session.userId, communityId)
       .then(isModerator => {
-        if (isModerator) {
+        if (isModerator || Admin.isSignedIn(req)) {
```
levity: this and the upper portion of the diff are suspiciously similar -- factor out into a function pls

api/controllers/NetworkController.js
levity: oh snap

api/controllers/NetworkController.js
```diff
@@ -1,3 +1,15 @@
+var updateCommunityIfModerator = (req, communityId, params, trx) =>
+  Membership.hasModeratorRole(req.session.userId, communityId)
+  .then(isModerator => {
+    if (isModerator || Admin.isSignedIn(req)) {
+      return Community.query()
+      .where('id', communityId)
+      .update(params)
+      .transacting(trx)
+    }
+    return
```
levity: this doesn't need to be here :)

api/controllers/NetworkController.js
```diff
@@ -1,3 +1,15 @@
+var updateCommunityIfModerator = (req, communityId, params, trx) =>
+  Membership.hasModeratorRole(req.session.userId, communityId)
+  .then(isModerator => {
```
levity: if you check Admin.isSignedIn first, you can skip the moderator role query entirely

PR #100, opened 2016-03-23
Post Controller returns comments and votes

robbiecarlton:


api/presenters/PostPresenter.js
```diff
@@ -1,24 +1,41 @@
...
+    {communities: qb => qb.column('community.id', 'name', 'slug', 'avatar_url')},
+    'contributions',
+    {'contributions.user': userColumns},
+    {followers: userColumns},
+    'media',
+    {relatedUsers: userColumns},
+    {responders: qb => qb.column('users.id', 'name', 'avatar_url', 'event_responses.response')},
+    (opts && opts.fromProject ? null : {projects: qb => qb.column('projects.id', 'title', 'slug')}),
+    (opts && opts.withComments
+      ? {comments: qb => qb.column('comment.id', 'text', 'created_at', 'user_id', 'post_id')}
```
robbiecarlton: ultimately this will return just the first 3 comments, but I thought I'd keep it like this for now for demoing the 'show all' on posts in the simplest way possible.

robbiecarlton:
@levity - Can i merge?

api/presenters/PostPresenter.js
```diff
@@ -62,10 +64,17 @@ var postAttributes = (post, opts) => {
...
+        {user: c.relations.user.pick('id', 'name', 'avatar_url')}
+      ))
   }
   if (opts.withVotes) {
-    console.log('VOTES', rel.votes)
+    extendedPost.voters = rel.votes.map(v => v.relations.user.pick('id', 'name', 'avatar_url'))
+
+    // recalculate myVote
+    // (the above calculation relies on rel.votes only containing the current users vote, which is not true if opt.withVotes)
+    extendedPost.myVote = _.includes(extendedPost.voters.map(v => v.id), userId)
```
levity: can you drop `votes` from the response here so that the new frontend code has to depend upon `voters.length`? then we can drop the nonstandard property once we phase out the old frontend

levity: or better yet, add it to the response in the first place only in an else-clause off of here

api/presenters/PostPresenter.js
```diff
@@ -62,10 +64,17 @@ var postAttributes = (post, opts) => {
       public: post.get('visibility') === Post.Visibility.PUBLIC_READABLE
     })
   if (opts.withComments) {
-    console.log('COMMENTS', rel.comments)
+    extendedPost.comments = rel.comments.map(c => _.merge(
+        c.pick('id', 'text', 'created_at', 'user'),
```
levity: indentation is off here

levity: likewise, drop `numComments` here or only add it in an else-clause

api/presenters/PostPresenter.js
```diff
@@ -13,25 +12,28 @@ var postRelations = (userId, opts) => {
     {responders: qb => qb.column('users.id', 'name', 'avatar_url', 'event_responses.response')},
     (opts && opts.fromProject ? null : {projects: qb => qb.column('projects.id', 'title', 'slug')}),
     (opts && opts.withComments
-      ? {comments: qb => qb.column('comment.id', 'text', 'user_id')}
+      ? {comments: qb => qb.column('comment.id', 'text', 'created_at', 'user_id', 'post_id')}
       : null),
     (opts && opts.withComments
       ? {'comments.user': userColumns}
       : null),
     (opts && opts.withVotes
```
levity: since you are setting the default value of `opts` to `{}`, you don't need to do `opts && ...`

api/presenters/PostPresenter.js
```diff
@@ -76,10 +85,10 @@ var postAttributes = (post, opts) => {
 // so that it knows when to stop expecting more.
 // we can't always use a naive approach to pagination, because
 // the order of results could shift while searching.
-var mapPresentWithTotal = function (posts, opts) {
+var mapPresentWithTotal = function (posts, userId, opts) {
```
levity: i would actually just drop this method and move the implementation into `PostController` -- since it's not used anywhere else, removing it would be one less layer of indirection

api/presenters/PostPresenter.js
robbiecarlton: I'd prefer to keep numComments for now, as we will likely soon not be returning the full list of comments with the post, in which case we'll still need the count.

levity: ah, yes, of course.

robbiecarlton:
@levity - how's this?

levity:
:+1:

PR #102, opened 2016-03-25
(WIP) Post tags

robbiecarlton:


robbiecarlton:
@levity - could you take a look at this.
All the functionality is done. Only reason it's still marked as WIP is that I'm getting some failing tests due to the posts_tags foreign constraint in the db. Which is weird because setup.clearDb is supposed to ignore those constraints. 
Wondering if you can see something I'm missing.
Also, let me know if there's anything you want me to change with the general approach

levity:
You need to set a specific property on the foreign key. Grep the old migrations for "DEFERRABLE".

robbiecarlton:
@levity - can you take a look at this when you get a chance. It would be great to merge it before you go away so I can do the frontend changes as well

api/models/Comment.js
```diff
@@ -25,6 +25,16 @@ module.exports = bookshelf.Model.extend({
 
   community: function () {
     return this.relations.post.relations.communities.first()
+  },
+
+  communities: function () {
+    this.load('post')
```
levity: this is an async method so you need `then` to be able to use it reliably

levity: but i guess you're not using this anymore?

robbiecarlton: right, that should be out

api/models/Tag.js
```diff
@@ -0,0 +1,105 @@
...
+  })
+  .then(tag => Promise.map(communities(taggable), com => addToCommunity(com, tag, taggable.user_id, trx)))
+}
+
+var removeFromTaggable = (taggable, tag, trx) => {
+  return taggable.tags().detach(tag.id, {transacting: trx})
+}
+
+var addToCommunity = (community, tag, user_id, trx) => {
+  return CommunityTag.where({community_id: community.id, tag_id: tag.id}).fetch()
```
levity: this needs to take place inside a transaction otherwise it's subject to a race condition

robbiecarlton: missed the save() there.
I can just pass trx to the save() call yeah? Or do I need to create a new transaction?

levity: ehh well actually it needs something like "SELECT ... FOR UPDATE", i.e. row locking... in lieu of that you'll want to have a (community_id, tag_id) uniqueness constraint on the table and be prepared for the save to fail -- the simple thing would be to assume that the save is failing due to the row having been created concurrently, and fetch again

migrations/20160324165302_tag_associations.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+exports.up = function (knex, Promise) {
+  return Promise.join(
+    knex.schema.createTable('posts_tags', function (table) {
+      table.increments()
+      table.bigInteger('post_id').references('id').inTable('post')
+      table.bigInteger('tag_id').references('id').inTable('tags')
```
levity: all of these join tables should have uniqueness constraints on their foreign key pairs

api/models/Tag.js
```diff
@@ -0,0 +1,105 @@
+var tagsInText = (text = '') => {
+  // TODO alphanumeric and underscore
+  return (text.match(/#\w+/g) || []).map(str => str.substr(1, str.length))
```
levity: you can omit the second argument to `substr` to go to the end of the string

robbiecarlton:
@levity - good to merge?

PR #107, opened 2016-04-06
Tag page

robbiecarlton:


api/models/Tag.js
```diff
@@ -15,7 +15,7 @@ var addToTaggable = (taggable, tagName, selected, trx) => {
     association = 'post.communities'
     communities = comment => comment.relations.post.relations.communities.models
   }
-  return taggable.load(association)
+  return taggable.load(association, {transacting: trx})
   .then(() => Tag.find(tagName))
   .then(tag => {
     if (tag) {
```
robbiecarlton: @levity there's a race condition here (if two people create the tag at the same time). Do you have any pointers on the best way to handle this/other places in the code where you've handled it?

levity: "catch the save error and find again" works in general -- i'm not totally sure how it behaves in this transaction context though, i.e. does the error surface immediately or only when you try to commit the transaction.

robbiecarlton: ok, I'll give it a try

On Wed, Apr 6, 2016 at 1:28 PM, Lawrence Wang <notifications@github.com>
wrote:

> In api/models/Tag.js
> <https://github.com/Hylozoic/hylo-node/pull/107#discussion_r58776054>:
>
> > @@ -15,7 +15,7 @@ var addToTaggable = (taggable, tagName, selected, trx) => {
> >      association = 'post.communities'
> >      communities = comment => comment.relations.post.relations.communities.models
> >    }
> > -  return taggable.load(association)
> > +  return taggable.load(association, {transacting: trx})
> >    .then(() => Tag.find(tagName))
> >    .then(tag => {
> >      if (tag) {
>
> "catch the save error and find again" works in general -- i'm not totally
> sure how it behaves in this transaction context though, i.e. does the error
> surface immediately or only when you try to commit the transaction.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/Hylozoic/hylo-node/pull/107/files/2ef4eadc29c40be40359d0862a3b8304b287601d#r58776054>
>



-- 
Robbie Carlton

719.358.3482
<robbie@authenticmanprogram.com>

robbiecarlton:
@levity - can I merge this?
Still working on the redux side, but this is a good place to merge for the backend

levity:
sure

PR #109, opened 2016-04-11
Catch unique constraint error

robbiecarlton:


robbiecarlton:
@levity - This fixes it. Potentially too general, what are your thoughts? Should I be checking for the specific error in the catch?

levity:
i would just change `updateForTaggable`:
```js
var toAdd = _.uniqBy(_.differenceBy(newTags, oldTags, 'id'), 'id'))
```

robbiecarlton:
Well that makes sense.
Ended up having to do some extra cleaning up.

PR #110, opened 2016-04-12
Tag following

robbiecarlton:


api/controllers/TagController.js
```diff
@@ -1,19 +1,91 @@
...
     .then(tag => {
       if (!tag) return res.notFound()
       return CommunityTag.where({community_id: res.locals.community.id, tag_id: tag.id})
-      .fetch({withRelated: 'owner'})
+      .fetch({withRelated: ['owner', 'community.followedTags']})
       .then(communityTag => {
         var result = communityTag.pick('id', 'description', 'community_id')
         result.name = tag.get('name')
         result.owner = communityTag.relations.owner.pick('id', 'name', 'avatar_url')
+        result.followed = !!communityTag.relations.community.relations.followedTags.find(ft => {
```
robbiecarlton: @levity - this find seems like something I should be doing in sql, but I'm not sure how to add those extra constraints. Is there a way to parameterize the bookshelf community.followedTags method to take user_id and tag_id?

robbiecarlton:
@levity - good to merge?

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
...
+      (tag, community) => {
+        if (!tag) return res.notFound()
+
+        return FollowedTag.where({
+          community_id: community.id,
+          tag_id: tag.id,
+          user_id: req.session.userId
+        }).fetch()
+        .then(followedTag => {
+          if (followedTag) {
```
levity: can this be a ternary expression please?

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
...
+    .catch(res.serverError)
+  },
+
+  findForLeftNav: function (req, res) {
+    return Community.find(req.param('communityId'))
+    .then(com => Promise.join(
+      fetchAndPresentFollowed(com.id, req.session.userId),
+      fetchAndPresentCreated(com.id, req.session.userId),
+      (followed, created) => ({
+        followed: _.filter(followed, f => !_.includes(_.map(created, 'name'), f.name)),
```
levity: i'm in favor of deprecating the global use of lodash and importing individual methods into a file's namespace like we do in hylo-redux. for one thing it's easier on the eyes

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
...
+        created: created
```
levity: use the "key and value are the same name" syntax here please

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
...
 
   findOne: function (req, res) {
     return Tag.find(req.param('tagName'))
     .then(tag => {
       if (!tag) return res.notFound()
       return CommunityTag.where({community_id: res.locals.community.id, tag_id: tag.id})
-      .fetch({withRelated: 'owner'})
+      .fetch({withRelated: [
+        'owner',
+        {'community.followedTags': qb => qb.where({'followed_tags.tag_id': tag.id, 'followed_tags.user_id': req.session.userId})}
```
levity: break this into multiple lines please

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
+var fetchAndPresentFollowed = (communityId, userId) =>
```
levity: this is exactly the same as `fetchAndPresentCreated` except for the class name

api/controllers/TagController.js
```diff
@@ -1,19 +1,90 @@
...
+      })
+    ))
+    .then(res.ok)
+    .catch(res.serverError)
+  },
+
+  follow: function (req, res) {
+    return Promise.join(
+      Tag.find(req.param('tagName')),
+      Community.find(req.param('communityId')),
```
levity: fetching the community object doesn't seem to be necessary here, since you only use its id

api/models/FollowedTag.js
```diff
@@ -0,0 +1,16 @@
+module.exports = bookshelf.Model.extend({
```
levity: can we call this TagFollow instead, in order to emphasize that the record represents an act of following, not a tag?

(feeling a similar weirdness about CommunityTag as well, but not sure what else to call it -- maybe TagCommunityMembership, which is long but consistent with what we're already doing e.g. PostProjectMembership)

api/controllers/TagController.js
robbiecarlton: and owner_id vs user_id.
Wasn't sure about refactoring, as it's not logically necessary that they're the same, they happen to be the same structure, but I can imagine that changing.
But sounds like you are for refactoring, so I'll do it.

api/controllers/TagController.js
robbiecarlton: I missed that we could do this on the backend. Great!

api/controllers/TagController.js
robbiecarlton: but req.param('communityId') is actually the slug, so I need to fetch for the actual id.
I called the param communityId by analogy with all the other routes that use it to mean id or slug.

levity: oh damn, right.

robbiecarlton:
@levity - how's this now?

api/controllers/TagController.js
```diff
@@ -1,19 +1,89 @@
...
+        return TagFollow.where({
+          community_id: community.id,
+          tag_id: tag.id,
+          user_id: req.session.userId
+        }).fetch()
+        .then(tagFollow => {
+          return tagFollow
+            ? tagFollow.destroy()
+            : new TagFollow({
+              community_id: community.id,
```
levity: indentation here should line up with `new`

levity:
we should get in the habit of using `const` rather than `var` where appropriate. otherwise looks good

api/controllers/TagController.js
robbiecarlton: I get a warning from the linter when I do that...

PR #112, opened 2016-04-18
route for posts in a tag across all communities

robbiecarlton:


robbiecarlton:
@levity - good to merge?

config/routes.js
```diff
@@ -95,6 +97,7 @@ module.exports.routes = {
   'POST   /noo/tag/:communityId/:tagName/follow':         'TagController.follow',
   'GET    /noo/tags/:communityId/followed':               'TagController.findFollowed',
   'GET    /noo/tags/:communityId/leftnav':                'TagController.findForLeftNav',
+  'GET    /noo/tag-in-all-communities/:tagName/posts':    'PostController.findForTagInAllCommunities',
```
levity: i should have caught this in the previous review. the `/noo/tag(s)/:communityId...` routes should be `/noo/community/:communityId/tags...` in keeping with the convention we've established where an ID in a URL is for the object named immediately preceding it. then this route becomes simply `/noo/tag/:tagName/posts`. please refactor.

api/controllers/PostController.js
```diff
@@ -97,6 +97,20 @@ const queryForTag = function (req, res) {
   }))
 }
 
+var queryForTagInAllCommunities = function (req, res) {
+  if (TokenAuth.isAuthenticated(res)) {
+    if (!RequestValidation.requireTimeRange(req, res)) return
```
levity: i think you want to drop these two if-clauses entirely? we don't need to support this janky token auth anymore and returning null here will cause an error for upstream code that expects a promise anyway.

robbiecarlton:
@levity made those changes, good to merge?

levity:
:cool:

PR #117, opened 2016-04-19
(WIP) make tag.find case insensitive

robbiecarlton:


api/models/Tag.js
```diff
@@ -97,7 +97,8 @@ module.exports = bookshelf.Model.extend({
   find: function (id, options) {
     if (!id) return Promise.resolve(null)
     if (isNaN(Number(id))) {
-      return Tag.where({name: id}).fetch(options)
+      return Tag.query(qb => qb.whereRaw('lower(name) = lower(?)', id))
```
robbiecarlton: Should I sanitize this?

levity: I think the prepared statement syntax will sanitize it already, but try it out in the console.

robbiecarlton: yup it does

PR #118, opened 2016-04-20
(WIP) Default tags

robbiecarlton:


api/models/Community.js
```diff
@@ -90,10 +90,26 @@ module.exports = bookshelf.Model.extend({
       return newPost.save({created_at: time, updated_at: time}, {transacting: trx})
       .then(() => Promise.all(_.flatten([
         self.posts().attach(newPost, {transacting: trx}),
+        Tag.updateForPost(newPost, newPost.get('type'), trx),
```
robbiecarlton: This line causes the method to break with an error about the transaction,

```
error: current transaction is aborted, commands ignored until end of transaction block
at [object Object].Connection.parseE (/Users/robbiecarlton/work/hylo/hylo-node/node_modules/pg/lib/connection.js:539:11)
```

test/unit/models/Membership.test.js
```diff
@@ -1,53 +1,72 @@
...
-
-    it('returns nothing for a blank user id', function() {
-      return Membership.find(null, community.id).then(function(membership) {
-        expect(membership).not.to.exist;
-      });
-    });
-
-    it('does not return an inactive membership', function() {
+          user.save(),
+          Tag.createDefaultTags()
```
levity: i think this is supposed to go in the `before` block for the `create` tests, not the `find` tests

api/models/Community.js
levity: not totally sure it's the root cause, but you have a few finds and loads downstream that don't use the transaction

robbiecarlton:
@levity adding the missing transaction fixed it, 
and I moved the default tag creation in that test.
Good to merge?

levity:
sweet. I wish the database error were a bit more descriptive. I guess the behavior is that if you make a database query without the transaction parameter, it closes the transaction first. anyway :shipit:

PR #119, opened 2016-04-21
return tag id in findOne

robbiecarlton:


api/controllers/TagController.js
```diff
@@ -58,9 +58,10 @@ module.exports = {
     .then(ct => {
       if (!ct) return res.notFound()
       return res.ok(merge(
-        ct.pick('id', 'description', 'community_id'),
+        ct.pick('description', 'community_id'),
         presentWithPost(tag),
         {
+          id: ct.get('tag_id'),
```
robbiecarlton: @levity - We're not currently using id, in tagsByCommunity, the tags are keyed by name. I was just returning id here out of habit, and maybe one day we'll need it.

Maybe I should just take id out of here for now?

levity:
let's keep returning it. i'm actually using its absence in TagPosts to determine whether fetchTag has completed or not.

levity:
i need this now so i'm merging it :)

PR #121, opened 2016-04-26
(WIP) Notifications refactor

robbiecarlton:


api/models/Notification.js
```diff
@@ -1,31 +1,15 @@
 module.exports = bookshelf.Model.extend({
-  tableName: 'notification',
 
-  actor: function() {
-    return this.belongsTo(User);
-  },
+  tableName: 'notifications',
```
robbiecarlton: @levity - someone has been here before me! Looks like this is old work so I'm clobbering it.

api/models/Activity.js
```diff
@@ -149,43 +155,65 @@ module.exports = bookshelf.Model.extend({
         trx))
   },
 
+  generateNotifications: function (activity) {
```
robbiecarlton: @levity - like this?

PR #124, opened 2016-05-12
Member cards

robbiecarlton:


api/controllers/UserController.js
```diff
@@ -248,10 +254,17 @@ module.exports = {
       }
     )
     var total
-
     Search.forUsers(options).fetchAll({withRelated: ['skills', 'organizations', 'memberships']})
     .tap(users => total = (users.length > 0 ? users.first().get('total') : 0))
     .then(users => users.map(u => UserPresenter.presentForList(u, res.locals.community.id)))
+    .then(list =>
```
robbiecarlton: @levity - would you take a look at how I'm getting the createdTags and the offerCount for members page here. I couldn't decide between this and putting it in to SearchForUsers / UserPresenter.
Let me know if you have a preference

robbiecarlton:
@levity good to merge?

api/controllers/UserController.js
levity: this is fine by me. note though, as i mentioned in slack, that the tags for a user should be the ones directly associated with them (tags_users), aka 'skills', not the ones associated with posts they've created

test/unit/services/Search.test.js
```diff
@@ -5,7 +5,7 @@ var time = require('time')
 
 describe('Search', function () {
   describe('.forPosts', function () {
-    it('produces the expected SQL for a complex query', function () {
+    it.skip('produces the expected SQL for a complex query', function () {
```
levity: remove?

api/services/tagPresentation.js
```diff
@@ -0,0 +1,31 @@
+export const fetchAndPresentTagJoins = (joinClass, communityId, userId) =>
```
levity: for consistency can we call this TagPresenter pls

robbiecarlton: the reason I didn't call it TagPresenter is because it's a collection of functions, not a class.
You want me to put the functions in a class called TagPresenter, or just rename the file TagPresenter?

levity: the other presenters are also just collections of functions

levity: just renaming the file is fine

robbiecarlton:
@levity - updated, can I merge this?

api/controllers/UserController.js
```diff
@@ -37,6 +37,11 @@ var setupReputationQuery = function (req, model) {
     }))
 }
 
+const countTaggedPosts = (userId, tag) =>
+  Search.forPosts({users: [userId], tag})
+  .fetchAll()
```
levity: on second view, this is actually really inefficient considering we're going to run this 20 times per page of members, because we're fetching all the supporting data for each post when all we want is the count.

this SQL gets us there:
```
select count(*), user_id from post p, posts_tags pt where p.id = pt.post_id and p.user_id in (21, 42) and pt.tag_id = 3 group by 2;
 count | user_id
-------+---------
     1 |      42
    52 |      21
```
or in knex:
```js
bookshelf.knex('post')
.join('posts_tags', 'post.id', 'posts_tags.post_id')
.where('post.user_id', 'in', [21, 42])
.where('posts_tags.tag_id', 3)
.groupBy('user_id')
.select(bookshelf.knex.raw('count(*), user_id'))
```

PR #126, opened 2016-05-18
remove projects from digest

robbiecarlton:


robbiecarlton:
@levity - merged this, check it out when you get the chance.

PR #127, opened 2016-05-20
Networks

robbiecarlton:


api/controllers/CommunityController.js
```diff
@@ -226,7 +226,7 @@ module.exports = {
         return Community.query(qb => {
           qb.where('network_id', network.get('id'))
           qb.where('community.active', true)
-          qb.select(bookshelf.knex.raw('community.id, community.slug, count(users_community.user_id) as "memberCount", count(community.id) over () as total'))
+          qb.select(bookshelf.knex.raw('community.id, community.slug, community.name, community.avatar_url, community.banner_url, count(users_community.user_id) as "memberCount", count(community.id) over () as total'))
```
levity: you can chain selects and each one will be added, so this would be better as:

```js
qb.select(bookshelf.knex.raw('count(...'))
qb.select('community.id', 'slug', ...) // omitting 'community' where possible
```

api/controllers/UserController.js
```diff
@@ -276,13 +276,13 @@ module.exports = {
...
-      communities: ids,
+    .then(communityIds => Search.forUsers({
+      communities: communityIds,
       limit: req.param('limit') || 20,
       offset: req.param('offset') || 0
-    }).fetchAll({withRelated: ['skills', 'organizations']}))
-    .tap(users => total = (users.length > 0 ? users.first().get('total') : 0))
-    .then(users => users.map(u => UserPresenter.presentForList(u)))
+    }).fetchAll({withRelated: ['skills', 'organizations', 'memberships']})
+      .tap(users => total = (users.length > 0 ? users.first().get('total') : 0))
```
levity: indentation

robbiecarlton: the `tap` and `then` are actually inside the block that starts `communityIds => `
Made a different change, see if it looks right to you now

levity: oh snap my mistake

PR #132, opened 2016-06-02
Topic notifications

robbiecarlton:


api/models/Tag.js
```diff
@@ -53,6 +53,12 @@ const addToCommunity = (community, tag, user_id, trx) => {
       user_id: user_id,
       created_at: new Date()
     }).save({}, {transacting: trx})
+    .then(() => new TagFollow({
```
robbiecarlton: a big change here: now when you create a tag you are automatically following it. This means we have new_post_counts for the created tags in the left nav.
You can still unfollow on the tag page, which just means your created but not followed tag will never be highlighted in the left menu.
This seemed preferable to adding a new_post_count to community tags and duplicating a lot of stuff.

robbiecarlton: I'll have to add tag_follows for existing community_tags in the db before merging this

robbiecarlton:
@levity - good to merge?

api/models/Tag.js
levity: makes sense to me -- this gives us the option to allow tag creators to unfollow straightforwardly as well

api/models/post/util.js
```diff
@@ -26,6 +26,16 @@ export const setupNewPostAttrs = function (userId, params) {
...
 }
 
+const updateTagFollows = (post, trxOpts) => post.load('tags', trxOpts)
+  .then(() => post.load('communities', trxOpts))
+  .then(() => TagFollow.query(q => {
+    q.whereIn('tag_id', post.relations.tags.map('id'))
+    q.whereIn('community_id', post.relations.communities.map('id'))
+  }).fetchAll(trxOpts))
+  .then(tagFollows =>
+    tagFollows.map(tagFollow =>
```
levity: this should be a Promise.map -- but better yet, this whole thing can be a single query. if you call `.query()` again above with no arguments, you'll drop down to the knex API, and then you can just call `.increment()`: http://knexjs.org/#Builder-increment

levity:
looks good apart from that change to `updateTagFollows`

PR #134, opened 2016-06-03
userController.findOne returns shared communities

robbiecarlton:


robbiecarlton:
@levity - good to merge?

api/presenters/UserPresenter.js
```diff
@@ -25,6 +25,16 @@ const recentTaggedPost = (userId, tag, viewingUserId) => {
   .then(post => post && PostPresenter.present(post, viewingUserId, opts))
 }
 
+var sharedCommunities = (user, viewingUserId) => Promise.join(
```
levity: we have `Membership.sharedCommunityIds` already which does this more efficiently, in a single query

PR #136, opened 2016-06-14
only count posts in community for mostActiveMembers, and use count instead of fetchAll

robbiecarlton:


robbiecarlton:
@levity - how's this?

api/services/TagPresenter.js
```diff
@@ -113,15 +118,15 @@ export const fetchAndPresentSummary = (community, tag) =>
...
+    .where({community_id: community.id, tag_id: tag.id})
+    .count(),
     mostActiveMembers(community, tag),
-    (communityTag, tagFollows, posts, activeMembers) => ({
+    (communityTag, followerCount, postCount, activeMembers) => ({
       description: communityTag.get('description'),
-      follower_count: tagFollows.length,
-      post_count: posts.length > 0 ? posts.first().get('total') : 0,
+      follower_count: followerCount,
+      post_count: postCount[0].count,
```
levity: you can wrap this in `Number` so the test can check for a number instead of a string

api/services/TagPresenter.js
```diff
@@ -113,15 +118,15 @@ export const fetchAndPresentSummary = (community, tag) =>
     CommunityTag.where({community_id: community.id, tag_id: tag.id})
     .fetch(),
     TagFollow.where({community_id: community.id, tag_id: tag.id})
-    .fetchAll(),
-    Search.forPosts({
-      communities: [community.id],
-      tag: tag.id
-    }).fetchAll(),
+    .count(),
+    bookshelf.knex('posts_tags')
```
levity: can you pop this out into a separate method for readability's sake? Maybe make it a class method in CommunityTag?

levity: you can also have the method do `.then(rows => Number(rows[0].count))`

levity:
looks good otherwise

PR #137, opened 2016-06-17
Join topic

robbiecarlton:


robbiecarlton:
@levity - please take a look when you have a chance

api/controllers/CommunityController.js
```diff
@@ -121,6 +121,7 @@ module.exports = {
 
   joinWithCode: function (req, res) {
     var community
+    var preexisting
```
levity: you can declare multiple vars on the same line when they're not assigned values

api/controllers/CommunityController.js
```diff
@@ -139,7 +141,27 @@ module.exports = {
...
+            tag_id: tag.id,
+            user_id: req.session.userId
+          }).save()
+        })
+        .catch(err => {
+          if (err.message && err.message.includes('duplicate key value')) {
+            return true
+          } else {
+            res.serverError(err)
+            return false
```
levity: can you write a test for this case? i think as written it will try to call `res.ok` after having called `res.serverError`

robbiecarlton: I actually can't think of a way of getting an error other than 'duplicate key value' there, so having a hard time writing the test ;)
Can you think of one? If not, I'll just remove this branch.

PR #141, opened 2016-06-29
Enable new digest

levity:


levity:
going to merge this after doing some testing in a production console to make sure we don't run out of memory

PR #142, opened 2016-07-01
add test for updating user settings

robbiecarlton:


robbiecarlton:
@levity - I couldn't reproduce the bug you're talking about. 
Should this test be covering it? If not, please tell me how to reproduce

levity:
oh, yeah, that's a good test for it. i realize now that i was overlooking the call to User#setSanely, which is what merges the JSON props instead of overwriting them. I guess the error I noticed was on the frontend... ? Anyway :+1:

PR #143, opened 2016-07-06
fix duplicate push notifications

robbiecarlton:


robbiecarlton:
@levity - merged this. take a look if you get a chance. Once it's live I'll clean up the db

api/controllers/DeviceController.js
```diff
@@ -28,15 +28,14 @@ module.exports = {
 
     return Device.forge({
```
levity: NBD, but Device.forge().fetch() looks weird; prefer Device.where().fetch() instead

levity:
:+1:

PR #147, opened 2016-07-14
Refactored tests to work in Australia.

rob-blasetti:


levity:
hi guys, thanks for creating the PR! i'll just make a few comments regarding style

test/setup/helpers.js
```diff
@@ -1,12 +1,23 @@
-export const spyify = (object, methodName, callback) => {
-  object['_original' + methodName] = object[methodName]
+const isSpy = (func) => {
+  return func.__spy
```
levity: great, thanks for making this test helper more robust! as far as style goes, i'd prefer a one-liner for such a simple function, and omit the parentheses around arguments when there's only one:
```js
const isSpy = func => !!func.__spy
```

test/unit/controllers/InvitationController.test.js
```diff
@@ -73,7 +73,7 @@ describe('InvitationController', () => {
       })
     })
 
-    it('sends invitations', function () {
+    it('sends invitations', () => {
```
levity: we do use ES6 function syntax wherever possible -- in this case we used the traditional syntax so that `this` in `this.timeout` would resolve correctly

test/unit/controllers/InvitationController.test.js
```diff
@@ -76,15 +73,12 @@ describe('InvitationController', () => {
...
 
-    it('sends invitations', function () {
-      this.timeout(10000)
+    it('sends invitations', () => {
       _.extend(req.params, {communityId: community.id, emails: 'foo@bar.com, bar@baz.com'})
 
       return InvitationController.create(req, res)
       .then(() => {
         expect(Email.sendInvitation).to.have.been.called.exactly(2)
-        expect(apiResult).to.exist
```
levity: though it's not really a "unit test", i was intentionally testing the interaction with the SendWithUs API here. i suppose i could give you a test API key for our account which just sends no email, because there are some other tests in the system as well that contact the SendWithUs API.

circle.yml
```diff
@@ -5,7 +5,7 @@ deployment:
   production:
     branch: master
     heroku:
-      appname: hylo-node
+      appname: hylo-node-hitfin
```
levity: hmm, i'm not going to want to merge this into master. Can you try setting it up as an environment variable or otherwise parameterizing it?

seeds/communities.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+exports.seed = function(knex, Promise) {
```
levity: nice 😎

seeds/communities.js
```diff
@@ -0,0 +1,11 @@
...
+    return Promise.join(
```
levity: note however that all promises passed to Promise.join are run concurrently, so you may want to do `del().then(() => Promise.join(...))` to avoid the occasional race condition

circle.yml
levity: btw, I confirmed with CircleCI support that using an environment variable here should work

test/setup/helpers.js
aaron-edwards: Yeah agreed, making the change now

test/unit/controllers/InvitationController.test.js
aaron-edwards: That is fair enough, the reason we went with this solution was because that section of the code that handles this is unit tested elsewhere. I am looking into another way of getting this test working for us now without mocking.

circle.yml
aaron-edwards: This actually wasn't suppose to be in the pull request. Good to know that there is an environment variable that we can use

seeds/communities.js
aaron-edwards: noted, will change also

rob-blasetti:
See PR [#154 ](https://github.com/Hylozoic/hylo-node/pull/154)

PR #156, opened 2016-08-03
Track signup events

robbiecarlton:
Goes with https://github.com/Hylozoic/hylo-redux/pull/92

### DIFF
--- a/api/services/Analytics.js
+++ b/api/services/Analytics.js
@@ -32,13 +32,16 @@ instance.pixelUrl = function (emailName, props) {
 }
 
 instance.trackSignup = function (userId, req) {
-  let event = 'Signup success'
+  let properties = {platform: 'Web'}
   if (req.headers['ios-version']) {
-    event += ' (iOS)'
+    properties.platform = 'ios'
   } else if (req.headers['android-version']) {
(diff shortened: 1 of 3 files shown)


robbiecarlton:
@levity - good to merge these two?

api/services/Analytics.js
```diff
@@ -31,4 +31,14 @@ instance.pixelUrl = function (emailName, props) {
   return prefix + encodedData
 }
 
+instance.trackSignup = function (userId, req) {
+  let event = 'Signup success'
+  if (req.headers['ios-version']) {
+    event += ' (iOS)'
```
levity: Don't change the event name; change the value of a property instead.

robbiecarlton: done

levity:
The important thing to test is that Mixpanel treats client- and server-side events as belonging to the same user. I think this is set up due to the way we make the `identify` call, but it may need to be repeated or done specifically after signup. The way to test this would be to look in Mixpanel at the stream of events for a test user, and confirm that an event before signup, the signup event, and an event after signup are all correctly assigned to that user.

levity:
I'm going to merge this locally and test what I described above

PR #157, opened 2016-08-04
For Slack: improve error catching and frontend service

Connoropolous:


api/services/Frontend.js
```diff
@@ -6,7 +6,7 @@ throughout the code.
 
 */
 
-const prefix = `${process.env.PROTOCOL}://${process.env.DOMAIN}`
+const prefix = `${process.env.PROTOCOL}://${process.env.FRONTEND_DOMAIN}`
```
levity: why did you find this to be necessary? DOMAIN is supposed to be the frontend domain already

Connoropolous: Ah. For some reason the other places where I saw it in use lead me to
believe that it was to represent the domain at which the server/api was
operating.

Connoropolous:
@levity simplified this to the only change that was truly needed

levity:
:cool:

PR #171, opened 2016-09-21
Admin default topics

robbiecarlton:


robbiecarlton:
@levity - will you take a look at this, see what you think.

levity:
... s/def/is_default/ ? :)

api/controllers/CommunityController.js
```diff
@@ -11,7 +11,7 @@ const welcomeMessage = 'Thank you for joining us here at Hylo. ' +
 
 module.exports = {
   find: function (req, res) {
-    Community
+    return Community
```
levity: why did you modify Community.find? as far as i can tell it's no longer used by the front-end

levity:
:cool:

PR #175, opened 2016-10-05
Feature/direct.messages

levity:
Continuation of #174

### DIFF
--- a/api/services/Search.js
+++ b/api/services/Search.js
@@ -69,9 +69,9 @@ module.exports = {
 
       if (!opts.type || opts.type === 'all') {
         qb.where(function () {
-          this.where('post.type', '!=', 'welcome')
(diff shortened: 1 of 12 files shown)


levity:
Hey @Connoropolous, re linter warnings: Robbie and I are using Atom and the atom-linter and linter-js-standard plugins. I imagine there's a vi equivalent, although if you wanted to switch to Atom that would be cool too.

api/services/Search.js
```diff
@@ -69,9 +69,9 @@ module.exports = {
 
       if (!opts.type || opts.type === 'all') {
         qb.where(function () {
-          this.where('post.type', '!=', 'welcome')
+          this.where('post.type', 'not in', ['welcome', Post.Type.THREAD])
```
levity: @Connoropolous this one's a common SQL gotcha -- if `post.type` is `null`, then `post.type != "thread"` is false.

PR #176, opened 2016-10-05
Request To Join Community

robbiecarlton:


robbiecarlton:
@levity - good to merge?

api/controllers/CommunityController.js
```diff
@@ -355,5 +360,76 @@ module.exports = {
...
     .catch(res.serverError)
+  },
+
+  requestToJoin: function (req, res) {
+    const { community } = res.locals
+    const params = {
+      community_id: community.id,
+      user_id: req.session.userId
+    }
+    if (!req.session.userId) return res.serverError(new Error('Unauthorized'))
```
levity: or, wait, just use that policy

api/models/Notification.js
```diff
@@ -86,6 +94,28 @@ module.exports = bookshelf.Model.extend({
     })
   },
 
+  sendJoinRequestPush: function () {
+    var communityIds = Activity.communityIds(this.relations.activity)
+    if (isEmpty(communityIds)) return Promise.resolve()
```
levity: we should probably throw an error here and in `sendApprovedJoinRequestPush` instead of doing nothing -- [fail fast](https://en.wikipedia.org/wiki/Fail-fast)

levity:
looks good otherwise

PR #177, opened 2016-10-12
follow created tags and return default with followed tags

robbiecarlton:


robbiecarlton:
@levity - good?

levity:
🆗 , though there's a race condition here if two people try to create a tag at the same time

robbiecarlton:
should I just put the whole create method in a transaction?

PR #178, opened 2016-10-13
Global new messages and new threads

Connoropolous:


api/models/Post.js
```diff
@@ -150,11 +156,32 @@ module.exports = bookshelf.Model.extend({
...
 
   pushTypingToSockets: function (userId, userName, isTyping, socketToExclude) {
-    this.sendToSubscribedSockets('userTyping', { userId, userName, isTyping }, socketToExclude)
+    var postId = this.id
+    this.sendToSubscribedSockets(`posts/${postId}`, 'userTyping', { userId, userName, isTyping }, socketToExclude)
+  },
+
+  pushSelfToSocket: function (userId) {
+    var that = this
+    const opts = { withComments: 'all' }
```
levity: style note: i've been putting spaces inside curly brackets only when they're being used for destructuring.

api/models/Post.js
```diff
@@ -150,11 +156,32 @@ module.exports = bookshelf.Model.extend({
...
   },
 
   pushTypingToSockets: function (userId, userName, isTyping, socketToExclude) {
-    this.sendToSubscribedSockets('userTyping', { userId, userName, isTyping }, socketToExclude)
+    var postId = this.id
+    this.sendToSubscribedSockets(`posts/${postId}`, 'userTyping', { userId, userName, isTyping }, socketToExclude)
+  },
+
+  pushSelfToSocket: function (userId) {
+    var that = this
```
levity: aliasing `this` isn't necessary with the version of node we're using

api/models/Post.js
```diff
@@ -150,11 +156,32 @@ module.exports = bookshelf.Model.extend({
...
 
   pushCommentToSockets: function (comment) {
-    this.sendToSubscribedSockets('commentAdded', comment)
+    var postId = this.id
+    this.sendToSubscribedSockets(`posts/${postId}`, 'commentAdded', comment)
+  },
+
+  pushMessageToSockets: function (message, userIds) {
+    var postId = this.id
+    const excludingSender = userIds.filter(id => id !== message.user_id.toString())
```
levity: toString() shouldn't be necessary here unless some funny business is happening upstream

api/models/Post.js
```diff
@@ -133,14 +140,13 @@ module.exports = bookshelf.Model.extend({
     return this.get('type') === Post.Type.WELCOME
   },
 
-  sendToSubscribedSockets: function (messageType, payload, socketToExclude) {
+  sendToSubscribedSockets: function (room, messageType, payload, socketToExclude) {
```
levity: this method looks ripe for extracting into a separate file in `api/services`

api/controllers/PostController.js
```diff
@@ -153,6 +153,16 @@ const checkPostTags = (attrs, opts) => {
   return throwErrorIfMissingTags(tags, opts.community_ids)
 }
 
+// called like responseForSockets(res)
+const responseForSockets = function (res) {
```
levity:
```js
const responseForSockets = res => err => err ? res.serverError(err) : res.ok({})
```
ok, i'm a one-liner junkie, it's true :)

levity: maybe call this `emptyResponse`

Connoropolous:
@levity made the suggested changes

api/services/Websockets.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function (room, messageType, payload, socketToExclude) {
```
levity: let's anticipate more functions ending up in this file, and export a named function instead of having it be the default export

PR #182, opened 2016-10-24
Updated Invite Page

robbiecarlton:


api/controllers/CommunityController.js
```diff
@@ -38,6 +38,21 @@ const afterCreatingMembership = (req, res, ms, community, preexisting) => {
   }))
 }
 
+const approveAJoinRequest = (req, res, joinRequest, community) => {
```
levity: no A in the name, please

levity: you could also refactor this to `curry((req, res, community, joinRequest) => ...)` and then use it below with the first three args filled out, e.g. `then(approveJoinRequest(req, res, community))`

api/controllers/InvitationController.js
```diff
@@ -125,5 +125,26 @@ module.exports = {
         })
       })
     .then(results => res.ok({results}))
+  },
+
+  reInviteAll: function (req, res) {
```
levity: reinvite is fine with a lowercase i

api/controllers/InvitationController.js
```diff
@@ -125,5 +125,26 @@ module.exports = {
...
+    const communityId = res.locals.community.id
+    return Invitation.where({community_id: communityId, used_by_id: null})
+    .fetchAll({withRelated: ['creator', 'community', 'tag']})
+    .then(invitations =>
+      Promise.map(invitations.models, invitation => {
```
levity: this should be queued because it could take quite a long time if there are lots of unsent invitations

api/controllers/InvitationController.js
```diff
@@ -125,5 +125,26 @@ module.exports = {
         })
       })
     .then(results => res.ok({results}))
+  },
+
+  reInviteAll: function (req, res) {
+    const communityId = res.locals.community.id
+    return Invitation.where({community_id: communityId, used_by_id: null})
```
levity: you also want to filter out invitations where used_by_id is null but there is someone with a matching email address, as we do in `InvitationController.find`

levity: if this is onerous (it is a pretty ugly query), an alternative would be to create an instance method on User that searches for an Invitation with matching email and marks it used, and then calling it during User.create

robbiecarlton:
Good to merge now?

api/models/User.js
```diff
@@ -220,7 +227,8 @@ module.exports = bookshelf.Model.extend(merge({
     return new User(attributes).save({}, {transacting: trx})
     .tap(user => Promise.join(
       account && LinkedAccount.create(user.id, account, {transacting: trx}),
-      community && Membership.create(user.id, community.id, {transacting: trx})
+      community && Membership.create(user.id, community.id, {transacting: trx}),
+      community && user.markInvitationsUsed(community.id)
```
levity: you need to pass through the transaction here

levity:
after that last fix you should be all set to merge

PR #185, opened 2016-11-03
latest message sender always has read the thread

Connoropolous:
this closes https://app.asana.com/0/155194938160003/206403127736287


Connoropolous:
You know what, don't merge this yet, I'm not fully satisfied with this

Connoropolous:
What I'm concerned about here is that 'creation' logic is getting blurred with 'viewed' logic. They might be distinct in a number of cases. Such as...

Person receives individual messages in their email inbox, and uses the email reply function to an old message and it updates their 'last viewed' value on the thread. doesn't seem right. I'm working on handling the 'last viewed' logic better on the front end

PR #189, opened 2016-11-23
Digest update

robbiecarlton:
This goes with the new Digest template created by Derek, (see the 'Simpler' version in SendWithUs).

I also need to add something in hylo-redux on the communities page for displaying a notification in the query string. (see line 214 of CommentController.js).

But nothing breaks if that's missing and I wanted to get this in first.

### DIFF
--- a/api/controllers/CommentController.js
+++ b/api/controllers/CommentController.js
@@ -196,6 +196,14 @@ module.exports = {
           failures = true
           return Promise.resolve()
         }
+        Analytics.track({
+          userId,
+          event: 'Post: Comment: Add by Email Form',
+          properties: {
+            post_id: post.id,
+            community: community && community.get('name')
+          }
+        })
         return createAndPresentComment(userId, replyText(post.id), post, {created_from: 'email batch form'})
       }))
     .then(() => {
--- a/api/controllers/CommentController.js
+++ b/api/controllers/CommentController.js
@@ -1,5 +1,5 @@
 import { difference, isEmpty, pickBy } from 'lodash'
-import { flow, filter, toPairs, map, includes } from 'lodash/fp'
+import { flow, filter, map, includes } from 'lodash/fp'
 import {
   handleMissingTagDescriptions, throwErrorIfMissingTags
 } from '../../lib/util/controllers'
@@ -181,9 +181,9 @@ module.exports = {
     const replyText = postId => markdown(req.param(`post-${postId}`))
 
     const postIds = flow(
-      toPairs,
-      filter(([k, v]) => k.match(/^post-(\d)+$/)),
-      map(([k, v]) => k.replace(/^post-/, ''))
+      Object.keys,
+      filter(k => k.match(/^post-(\d)+$/)),
+      map(k => k.replace(/^post-/, ''))
     )(req.allParams())
 
     var failures = false
@@ -198,7 +198,7 @@ module.exports = {
         }
         return createAndPresentComment(userId, replyText(post.id), post, {created_from: 'email batch form'})
       }))
-    .then(post => {
+    .then(() => {
       var notification
       if (failures) {
         notification = 'Some of your comments could not be added.'
--- a/api/controllers/CommentController.js
+++ b/api/controllers/CommentController.js
@@ -4,7 +4,7 @@ import {
   handleMissingTagDescriptions, throwErrorIfMissingTags
 } from '../../lib/util/controllers'
 import { normalizeComment } from '../../lib/util/normalize'
-import { sanitize } from 'hylo-utils/text'
+import { sanitize, markdown } from 'hylo-utils/text'
 
 const userColumns = q => q.column('id', 'name', 'avatar_url')
 
@@ -176,15 +176,9 @@ module.exports = {
   },
 
   createBatchFromEmailForm: function (req, res) {
-    try {
-      var tokenData = Email.decodeFormToken(req.param('token'))
-    } catch (e) {
-      return Promise.resolve(res.serverError(new Error('Invalid token: ' + req.param('token'))))
-    }
+    const { communityId, userId } = res.locals.tokenData
 
-    const { communityId, userId } = tokenData
-
-    const replyText = postId => req.param(`post-${postId}`)
+    const replyText = postId => markdown(req.param(`post-${postId}`))
 
     const postIds = flow(
       toPairs,
--- a/api/controllers/CommentController.js
+++ b/api/controllers/CommentController.js
@@ -1,4 +1,5 @@
 import { difference, isEmpty, pickBy } from 'lodash'
+import { flow, filter, toPairs, map, includes } from 'lodash/fp'
 import {
   handleMissingTagDescriptions, throwErrorIfMissingTags
 } from '../../lib/util/controllers'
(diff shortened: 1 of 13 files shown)


robbiecarlton:
Oh, it has the database cleanup branch in it (because my DB is now in the new structure)
Everything that's new in this one is in the final commit "add route for creating multiple comments from the digest email form"

test/unit/controllers/CommentController.test.js
```diff
@@ -216,4 +216,42 @@ describe('CommentController', function () {
       })
     })
   })
+
+  describe.only('createBatchFromEmailForm', () => {
```
levity: remove `only` pls

levity: btw, in dev, you can do `grunt autotest --file=test/unit/controllers/CommentController.test.js` to just re-run that one file

levity: similarly e.g. `npm run autotest -- -f test/client/components/Login.test.js` for hylo-redux

api/controllers/CommentController.js
```diff
@@ -172,5 +173,45 @@ module.exports = {
...
+  createBatchFromEmailForm: function (req, res) {
+    try {
+      var tokenData = Email.decodeFormToken(req.param('token'))
+    } catch (e) {
+      return Promise.resolve(res.serverError(new Error('Invalid token: ' + req.param('token'))))
+    }
+
+    const { communityId, userId } = tokenData
+
+    const replyText = postId => req.param(`post-${postId}`)
```
levity: `createAndPresentComment` expects to get reply text that is properly formatted HTML (which it does from TinyMCE), otherwise it's not going to display correctly on the site and in email notifications. the simplest way is to parse the text as markdown; see e.g. `Comment.cleanEmailText`

api/controllers/CommentController.js
```diff
@@ -172,5 +173,45 @@ module.exports = {
       })
       .then(res.ok, res.serverError)
     })
+  },
+
+  createBatchFromEmailForm: function (req, res) {
+    try {
```
levity: this would be a good thing to pull into a policy

api/controllers/CommentController.js
```diff
@@ -172,5 +173,45 @@ module.exports = {
...
+      var tokenData = Email.decodeFormToken(req.param('token'))
+    } catch (e) {
+      return Promise.resolve(res.serverError(new Error('Invalid token: ' + req.param('token'))))
+    }
+
+    const { communityId, userId } = tokenData
+
+    const replyText = postId => req.param(`post-${postId}`)
+
+    const postIds = flow(
```
levity: this is confusing me, partly because i'm used to seeing `filter((v, k) => ...)`. if all you want is the post ids, i would do: 
```js
flow(
  Object.keys,
  filter(k => k.match(/^post-(\d)+$/)),
  map(k => k.replace(/^post-/, ''))
)(req.allParams())

File: config/routes.js [hunk 5]
```diff
@@ -117,6 +117,7 @@ module.exports.routes = {
   'POST   /noo/hook/comment':                             'CommentController.createFromEmail',
   'POST   /noo/hook/message':                             'MessageController.relayFromEmail',
   'POST   /noo/hook/postForm':                            'PostController.createFromEmailForm',
+  'POST   /noo/hook/replyForm':                           'CommentController.createBatchFromEmailForm',
```
levity: maybe call this `commentForm` or `batchCommentForm` for consistency's sake?

lib/community/digest2/index.js
```diff
@@ -39,10 +39,13 @@ export const prepareDigestData = (id, type, opts = {}) => {
 }
 
 const sendToUser = (user, data, opts = {}) =>
-  personalizeData(user, data, opts).then(data =>
+  personalizeData(user, data, opts)
+  .tap(data => console.log('sending digest with data', data))
```
levity: remove this console.log

lib/community/digest2/personalizeData.js
```diff
@@ -82,8 +82,9 @@ const personalizeData = (user, data, opts = {}) => {
     },
     email_settings_url: Frontend.Route.userSettings() + clickthroughParams + '&expand=account',
     tracking_pixel_url: Analytics.pixelUrl('Digest', {userId: user.id, community: data.community_name}),
-    form_action_url: Frontend.Route.emailPostForm(),
```
levity: you can't rename this; it's still used by the other template version. well, you can, but you'll have to edit the other template version once this is deployed. if you want to do that, i would suggest creating a subtask in Asana for it.

lib/community/digest2/index.js
```diff
@@ -39,10 +39,13 @@ export const prepareDigestData = (id, type, opts = {}) => {
...
+  .then(data =>
     opts.dryRun ||
     Email.sendSimpleEmail(user.get('email'), templateId, data, {
-      sender: {name: data.community_name || 'Hylo'}
+      sender: {name: data.community_name || 'Hylo'},
+      version_name: 'simpler'
```
levity: wait a minute, we're not AB testing this email design?

levity:
Two more general concerns:

1) The "simpler" template is set up with a placeholder subject line, "User and X others are looking for help" -- is that supposed to be implemented here?

2) The "simpler" template only has sections for new offers and requests. But there is no change to the data-gathering code in this PR. So if the data for a community consists of only new comment, or new posts that aren't offers or request, then users will receive an empty email. (Technically they'd receive an email containing only the post prompt, but that's not supposed to happen either, except once a week for communities that have `settings.post_prompt_day` set.)

lib/community/digest2/personalizeData.js
robbiecarlton: I've created an Asana subtask to rename. I was thinking we were switching not AB testing, so wasn't an issue.

robbiecarlton:
Re your concerns:
1) Yes, that's supposed to be implemented here, missed it in the rush yesterday. I'll add it.

2) "simpler" template also has a section for "conversations" which I think are all posts that aren't requests or offers. 
And with comments, it doesn't send a blank email, it sends an email with the posts that the comments are replies to, but without the comments. So yes, it's broken. 

Couple of options: I could re add the comments to this template, or 
drop comments from digest altogether, which I think you'd suggested doing anyway.

3 alternative is to let the digest generation code be aware of the AB testing choice, but seems like we'd want to avoid this if possible.

levity:
Oh, the conversations are still in the template... In that case, the template layout is confusing because there's no header between offers and conversations.

levity:
I also realize now that dropping the post description from the template is not going to work. If you sample recent post titles, they're simply not explanatory enough on their own.

api/controllers/CommentController.js
```diff
@@ -172,5 +173,39 @@ module.exports = {
...
+
+    return Community.find(communityId)
+    .then(community => Promise.map(postIds, id =>
+      Post.find(id, {withRelated: ['communities']})
+      .then(post => {
+        if (!post || !includes(communityId, post.relations.communities.pluck('id'))) {
+          failures = true
+          return Promise.resolve()
+        }
+        return createAndPresentComment(userId, replyText(post.id), post, {created_from: 'email batch form'})
```
levity: We need to track this event in analytics like we do for `createFromEmail`

robbiecarlton:
@levity - good to merge now?

lib/community/digest2/personalizeData.js
```diff
@@ -81,9 +81,10 @@ const personalizeData = (user, data, opts = {}) => {
       name: user.get('name')
     },
     email_settings_url: Frontend.Route.userSettings() + clickthroughParams + '&expand=account',
-    tracking_pixel_url: Analytics.pixelUrl('Digest', {userId: user.id, community: data.community_name}),
-    form_action_url: Frontend.Route.emailPostForm(),
-    form_token: Email.postCreationToken(data.community_id, user.id)
+    tracking_pixel_url: Analytics.pixelUrl('Digest', {userId: user.id, community: data.community_name, versionName: opts.versionName}),
```
levity: can you make the key "Email Version" instead of "versionName"? The key shows up in Mixpanel exactly as it's defined here.

robbiecarlton: done!
I can merge?

levity:
:ok:

PR #190, opened 2016-12-01
token based API access

Connoropolous:
context and questions:
- token can be provided in a `POST` as `access_token` in the `req.body`, as `access_token` as a query parameter, or in the headers as `x-access-token`
- @levity is it ok that we clean up the old token auth stuff that was just for Harlan? I've done that here
- `accessTokenAuth` is a new policy that can be added to any route. it should always be added BEFORE `sessionAuth`, but could be used without `sessionAuth`

screen shot 2016-12-01 at 3 51 27 pm
screen shot 2016-12-01 at 3 51 50 pm


Connoropolous:
goes with https://github.com/Hylozoic/hylo-redux/pull/177

levity:
Reviewing this now. I'm not sure about this change from `req.session.userId` to `req.getUserId`. I think it could actually be a useful feature in some cases to set `req.session.userId` based on successful token auth, and in any case it makes this diff much smaller.

If we were going to do a big refactoring around `req`, I'd be interested in getting closer to the conventions of [passport](http://passportjs.org/), which normally sets `req.user`. The only reason we don't do it that way already is because of legacy reasons which are no longer relevant.

levity:
More detail about passport: We have it set up so that for admin login, we do passport's normal flow and call `req.login`, which sets `req.user` (see `AdminSessionController`). And for normal user login, we do `UserSession.login`. Ideally we'd use passport's normal flow for normal user login and only do something special for admins. (*if* we wanted to do this refactoring--it's not a huge priority)

api/models/Token.js
```diff
@@ -0,0 +1,11 @@
+module.exports = bookshelf.Model.extend({
+  tableName: 'tokens',
```
levity: instead of creating a new model, we could also provide the same functionality by using the existing `LinkedAccount` model, setting its `provider_key` value to 'token' and `provider_user_id` to the token's value.

api/services/TokenAuth.js
```diff
@@ -1,21 +1,29 @@
...
+  generateToken: function() {
+    const randomBytes = Promise.promisify(crypto.randomBytes)
+    return randomBytes(24).then(buffer => buffer.toString('hex'))
   },
 
-  isAuthenticated: function(res) {
-    return !!res.locals.tokenAuthenticated;
+  checkAndSetAuthenticated: function(req) {
+    req.body = req.body || {}
+    const token = req.body.token || req.query.token || req.headers['x-access-token']
```
levity: let's call the param `access_token` or `auth_token` throughout. we already use the name `token` for other purposes, e.g. in `InvitationController`.

config/routes.js
```diff
@@ -135,6 +135,9 @@ module.exports.routes = {
   'GET    /noo/logout':                                   'SessionController.destroy',
   'DELETE /noo/session':                                  'SessionController.destroySession',
 
+  'POST   /noo/token':                                    'TokenController.create',
```
levity: let's also use the term `access_token` or `auth_token` in the routes here and in the controller name

Connoropolous:
"I think it could actually be a useful feature in some cases to set req.session.userId based on successful token auth, and in any case it makes this diff much smaller."

@levity the only downside to this is that its misleading because the userId isn't being set in relation to a session, but to a token.

levity:
I meant that it could be useful in some cases to set the user ID in the
session based on a token, just as you would with a password.

Connoropolous:
@levity I made all those changes... the way I have it now, one is able to 'login' and set the session by using the their token

api/services/AccessTokenAuth.js
```diff
@@ -0,0 +1,35 @@
...
+       qb.where('linked_account.provider_user_id', '=', token)
+       qb.andWhere('linked_account.provider_key', '=', 'token')
+    })
+    .fetch()
+    .then(user => {
+      if (user) {
+        req.session = req.session || {}
+        req.session.authenticated = true
+        req.session.userId = user.id
+        req.session.tokenAuthenticaed = true
```
levity: typo here

api/policies/accessTokenAuth.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function(req, res, next) {
+  AccessTokenAuth.checkAndSetAuthenticated(req).then(() => next())
```
levity: add a `.catch(err => ...)` with `res.serverError` or something like that, otherwise this could hang the response

api/services/AccessTokenAuth.js
```diff
@@ -0,0 +1,35 @@
...
+  generateToken: function() {
+    const randomBytes = Promise.promisify(crypto.randomBytes)
+    return randomBytes(24).then(buffer => buffer.toString('hex'))
+  },
+
+  checkAndSetAuthenticated: function(req) {
+    req.body = req.body || {}
+    const token = req.body.access_token || req.query.access_token || req.headers['x-access-token']
+    if (!token) return Promise.resolve()
+    return User.query(function (qb) {
```
levity: you don't actually need a full User object here, so you could do `LinkedAccount.where(...).fetch().then(account => account.get('user_id'))`

api/policies/sessionAuth.js
```diff
@@ -8,7 +8,7 @@ var fail = function (res) {
 }
 
 module.exports = function (req, res, next) {
-  if (TokenAuth.isAuthenticated(res)) {
+  if (AccessTokenAuth.isAuthenticated(req)) {
```
levity: this if-clause can be dropped entirely; since `req.session.userId` is set now, the `UserSession.isLoggedIn` check will be true

api/services/AccessTokenAuth.js
levity: actually, scratch that -- better to call `UserSession.login` with the user object, otherwise `UserSession.isLoggedIn` won't behave as expected

levity: check out the implementation of `UserSession.login` and maybe add a new argument if you don't want API token use to increment `last_login` on the user

api/controllers/AccessTokenController.js
```diff
@@ -0,0 +1,22 @@
+const Promise = require('bluebird')
+
+module.exports = {
+  create: function (req, res) {
+    return LinkedAccount.tokenForUser(req.session.userId)
+    .then(account => account ? Promise.reject(new Error('User already has a token generated')) : Promise.resolve())
```
levity: small quibble -- you could do `account || Promise.reject(...)` instead of a conditional expression here

levity:
looking good; once you make the changes i suggested above, we should be good to go

Connoropolous:
K! On it

api/controllers/PostController.js
```diff
@@ -58,10 +58,6 @@ const findTagId = req =>
   req.param('tag') && Tag.find(req.param('tag')).then(t => t.id)
 
 const queryForCommunity = function (req, res) {
-  if (TokenAuth.isAuthenticated(res)) {
-    if (!RequestValidation.requireTimeRange(req, res)) return
```
Connoropolous: @levity it's ok to take this out right? it's part of the specific use case with harlan?

levity: Yes and no. This was something I put in place not only because Harlan asked for it, but because it would avoid the default behavior being "fetch everything for a community", to reduce impact on the server. We don't have to do it this way, but it is something we should think about. Maybe a default value for `limit` would do as well.

Connoropolous: Got it. I can do the default limit thing here

PR #191, opened 2016-12-06
GraphQL

levity:
- [x] express request handler
- [x] easy way of mapping bookshelf models to graphql response
- [x] avoid N+1 problem
- [x] filter data based on what the user should have access to


levity:
@Connoropolous here's what i've got so far

Connoropolous:
very cool! thanks for posting. 
so graphql then will/can potentially handle any part of the API related to fetching data for use by a client?

Connoropolous:
I wonder if other people could also benefit from the graphql-bookshelf adapter, if we get it working well

levity:
Yep! And making changes too, once I figure out how that works.

levity:
Yes, I think so. I've taken a different approach than another project that I found out in the wild, by still hand-crafting a schema using GraphQL's schema definition language, and then supplying only its implementation in Javascript. It's too early to say whether it's a good approach overall, but so far I like how the standalone schema file makes a good API reference "cheat sheet".

levity:
hey guys, just wanted to get some eyes on this -- if you like, i can take some time tomorrow to walk you all through it on video.

levity:
it occurs to me that it may be helpful for me to write some tests first :)

Connoropolous:
@levity how would you feel about allowing AccessTokenAuth on the graph endpoint. That would be so powerful and make so much possible!

levity:
@Connoropolous yes, we can probably do that.

lorenjohnson:
Excellent work Lawrence! The code looks clear enough and the SOC makes sense to me.  

It could be cool to open-source the GraphQL Bookshelf adapter piece of this eventually.

I'm looking forward to getting into this more, but as long as this mounts on express without errors and the app builds then I see no risk in the merging of this code now. 

I'll get a much better sense for things once I utilize this API on my next feature build, or in seeing you set things up for "Popular Skills". 

LGTM👍

PR #192, opened 2016-12-06
Reply to comment

robbiecarlton:


api/controllers/CommentController.js
```diff
@@ -20,40 +20,60 @@ const presentComment = (comment) =>
     return Object.assign(buckets, c)
   })
 
-const createAndPresentComment = function (commenterId, text, post, opts = {}) {
+const createAndPresentComment = function (commenterId, text, post, parentComment, opts = {}) {
```
levity: general design guideline: if you go past two or so args, start putting them into an options object. reading the argument names is much easier to follow than remembering arguments by their order. so in this case i would move `parentComment` into `opts`

levity: general design guideline: if you go past two or so args, start putting them into an options object. reading the argument names is much easier to follow than remembering arguments by their order. so in this case i would move `parentComment` into `opts`

levity:
for the sake of posterity, and for guiding my review, can you write up a short bullet list of the features this is designed to support? e.g. "when you reply to a comment, you are added to the list of followers for the comment, but not the post", or whatever it is.

knexfile.js
```diff
@@ -1,4 +1,4 @@
-import { merge } from 'lodash'
+const merge = require('lodash').merge
```
levity: i'm assuming you did this because the migrations script wasn't running babel? my bad. you can also require lodash functions individually, e.g.: `const merge = require('lodash/merge')`

levity:
actually, never mind, it's basically clear enough from the tests.

api/models/Comment.js
```diff
@@ -39,12 +40,33 @@ module.exports = bookshelf.Model.extend({
...
-    return this.load(['post', 'post.followers'])
+    const isReplyToPost = !this.get('comment_id')
+
+    var toLoad = ['post', 'post.followers']
+    if (!isReplyToPost) toLoad = toLoad.concat(['comment', 'comment.followers'])
+
+    return this.load(toLoad)
     .then(() => {
-      const followers = this.relations.post.relations.followers.map(follower => ({
+      const followers = (isReplyToPost
```
levity: `this.relations[isReplyToPost ? 'post' : 'comment'].relations.followers` is a bit more concise

api/models/post/util.js
```diff
@@ -172,3 +172,32 @@ export const createThread = (userId, params) =>
...
+    .tap(follow => {
+      if (!createActivity) return
+
+      var updates = []
+      const addActivity = (recipientId, method) => {
+        updates.push(Activity[method](follow, recipientId)
+        .save({}, _.pick(opts, 'transacting'))
+        .then(activity => activity.createNotifications(transacting)))
+      }
+      if (followerId !== addedById) addActivity(followerId, 'forFollowAdd')
```
levity: this is not really related to the PR, but I think we can ditch these two activity types.

migrations/schema.sql
```diff
@@ -1365,10 +1368,8 @@ UNION
  SELECT NULL::bigint AS post_id,
     u.id AS user_id,
     NULL::bigint AS comment_id,
-    ((setweight(to_tsvector('english'::regconfig, (u.name)::text), 'A'::"char") || setweight(to_tsvector('english'::regconfig, COALESCE(string_agg(replace((t.name)::text, '-'::text, ' '::text), ' '::text), ''::text)), 'C'::"char")) || setweight(to_tsvector('english'::regconfig, COALESCE(u.bio, ''::text)), 'C'::"char")) AS document
```
levity: this shouldn't be part of the diff

knexfile.js
robbiecarlton: yep. Cool, I'll use that way it's nicer.

api/models/post/util.js
robbiecarlton: agreed, and hallelujah. I'll create a new task for this because it touches a few places

migrations/schema.sql
levity: yes. it means the version of the search_index materialized view in your local db doesn't match the current one. you can always `FullTextSearch.dropView().then(() => FullTextSearch.createView())` to bring it up to date.

robbiecarlton:
@levity - Good to merge now?

PR #194, opened 2016-12-10
Complete Request Flow - Backend change

lorenjohnson:
NOTE: Must be released in concert with: https://github.com/Hylozoic/hylo-redux/pull/182

* Adds  Activity/Notifications for contributors who have been marked as contributing to a request
* Removes Contributors upon unfulfilling (uncompleting) a #request Post
* Updates the name of the key for Contributors on a Post
* Splits Post "util" functions and Post "request type" functions both into a models/post subdirectory in their own files for better separation of concerns

Add'l FYI:

* schema.sql: I had some trouble getting the knex migrations to run from scratch though I found a way for it to mostly work. As-is though I don't trust my schema.sql to be entirely correct so have not checked it in.
* Add'l testing: The Contribution model could still use a basic unit test or two for #createActivities
* I have not been able to get the spys working on the Post#fulfillRequest tests for Queue.classMethod, I've for now commented-out the relevant tests...


api/controllers/PostController.js
```diff
@@ -322,14 +322,12 @@ const PostController = {
...
-    return bookshelf.transaction(trx =>
-      post.save({fulfilled_at}, {patch: true, transacting: trx})
-      .tap(() => Promise.map(contributorIds, userId =>
-        Contribution.create(userId, post.id, trx))))
-    .then(() => res.ok({}))
+    const contributorIds = req.param('contributorIds') || []
+    const fulfilledAt = post.get('fulfilled_at')
+    const result = fulfilledAt ?
+      post.unfulfillRequest() :
+      post.fulfillRequest({ contributorIds })
```
levity: style quibble: the rule i've been following regarding spaces on the insides of curly brackets is that i only do it for destructuring.

lorenjohnson: SGTM: so to make sure I'm understanding: Do you mean wrapping in spaces for destructing in the case of assignmean only? So to follow this on line 329 I'd remove the spaces around { contributorIds } to make {contributorIds}?

api/models/Contribution.js
```diff
@@ -7,12 +7,33 @@ module.exports = bookshelf.Model.extend({
...
-  create: (user_id, post_id, transacting) =>
-    new Contribution({post_id, user_id, contributed_at: new Date()})
-    .save(null, {transacting}),
+  find: (id, options) => Contribution.where({id}).fetch(options),
+
+  create: function(user_id, post_id, trx) {
+    return new Contribution({post_id, user_id, contributed_at: new Date()})
+    .save(null, {transacting: trx})
+    .then((contribution) => {
+      Queue.classMethod('Contribution', 'createActivities',  {
```
levity: need to return the promise here

api/controllers/PostController.js
levity: Yep. This case isn't destructuring, it's a shorthand for object literals. (Don't know what it's called.)

test/setup.js
```diff
@@ -10,6 +10,7 @@ import nock from 'nock'
 
 chai.use(require('chai-spies'))
 chai.use(require('chai-as-promised'))
+chai.use(require('chai-datetime'));
```
levity: no semicolon please :)

test/unit/models/post/request.test.js
```diff
@@ -0,0 +1,81 @@
...
+        fulfilledAt,
+        contributorIds: [contributor1.id, contributor2.id]
+      })
+      .then(post => post.fetch({withRelated: 'contributions'}))
+    )
+
+    it('should remove fulfilled time', () => {
+      expect(post.get('fulfilled_at')).to.equalDate(fulfilledAt)
+      return post.unfulfillRequest().then(() =>
+        expect(post.get('fulfilled_at')).to.not.exist
```
levity: another style quibble -- for one-liners like this, i like to put the closing `)` on the same line as the statement -- i think it makes the one-liner more visually differentiated so we can more quickly determine that `return` isn't needed

test/unit/models/post/request.test.js
```diff
@@ -0,0 +1,81 @@
...
+      expect(post.get('fulfilled_at')).to.equalDate(fulfilledAt)
+    })
+
+    it('should add contributors', () => {
+      expect(post.relations.contributions).to.be.length(2)
+      expect(post.relations.contributions.map((c) => c.get('user_id')))
+        .to.include.members([contributor1.id, contributor2.id])
+    })
+
+    // it('should add activities and notifications to contributors', () => {
```
levity: why is this commented out?

lorenjohnson: The spy doesn't see the calls in the test, though direct logging shows that the Queuing does get called during the test. Happy to diagnose this further but I'd like you to look at it with me and see if you can see what I'm missing. Didn't want to hold the review upon this one issue any longer...

api/models/Contribution.js
lorenjohnson: Perhaps this has something to do with why I couldn't get the spy'd Queue.classMethod to be noticed in my commented-out test below. Checking...

lorenjohnson: (Fixed, but unfortunately didn't help said issue)

test/unit/models/post/request.test.js
lorenjohnson: 1) post/request #fulfillRequest should add activities and notifications to contributors:

     AssertionError: expected { Spy, 2 calls } to have been called with [ Array(3) ]

lorenjohnson: This is now fixed. I was looking at the contributor.id instead of the actual *contribution*.id

levity:
nice! feel free to merge. (style quibble: no need for parens around single arguments in fat-arrow functions. but don't let that hold up the merge.)

PR #197, opened 2016-12-15
Return Popular Skills with community in findOne

robbiecarlton:


robbiecarlton:
Yes, I debated including it in its own api call.
Looking forward to seeing what you do with the GraphQL integration

PR #199, opened 2016-12-22
Track logins from mobile apps

robbiecarlton:


api/services/UserSession.js
```diff
@@ -8,6 +10,20 @@ module.exports = {
     req.session.version = this.version
 
     if (providerKey === 'admin') return
+
+    if (req.headers['ios-version'] || req.headers['android-version']) {
+      const properties = omitBy(isNil, {
+        iosVersion: req.headers['ios-version'],
```
levity: Same deal here as with the other PR: "App Version (iOS)" and "App Version (Android)"

PR #201, opened 2016-12-29
Resend unused Invites

robbiecarlton:


api/models/Invitation.js
```diff
@@ -85,32 +100,40 @@ module.exports = bookshelf.Model.extend({
...
-          message: markdown(message),
-          moderator,
-          subject
-        }
-        return invitation.send(opts)
-        .then(() => invitation.save({created_at: new Date()}, {patch: true}))
-      }))
+      Promise.map(invitations.models, invitation =>
+        invitation.send()
+        .then(() => invitation.save({created_at: new Date()}, {patch: true}))))
```
levity: don't change `created_at` here

test/unit/models/Invitation.test.js
```diff
@@ -107,4 +112,161 @@ describe('Invitation', function () {
...
+            email: 'd@b.com',
+            sent_count: 4
+          }
+        ])
+
+        expect(sortBy('email', invitations.map(i => ({
+          email: i.get('email'),
+          sent_count: i.get('sent_count')
+        })))).to.deep.equal(expected)
+      })
```
levity: let's also check the value of `last_sent_at` here -- you can use [closeTo](http://chaijs.com/api/bdd/#method_closeto)

robbiecarlton:
I already check that sent_count doesn't change. Will check the last_sent_dates

robbiecarlton:
good to merge?

PR #202, opened 2016-12-31
Add graphql getters for Community memberCount and postCount

robbiecarlton:


api/models/Community.js
```diff
@@ -146,6 +146,28 @@ module.exports = bookshelf.Model.extend(merge({
     })
     .fetchAll()
     .then(skills => skills.pluck('name'))
+  },
+
+  memberCount: function () {
+    return User.query(q => {
```
levity: it really rankles that there isn't a good way to do this based on the `hasMany` methods we've already defined.

api/models/Community.js
```diff
@@ -146,6 +146,28 @@ module.exports = bookshelf.Model.extend(merge({
...
+      q.select(bookshelf.knex.raw('count(*)'))
+      q.join('communities_users', 'users.id', 'communities_users.user_id')
+      q.where('communities_users.community_id', this.id)
```
levity: you can do both of these in a single `where` if you pass it an object literal

api/models/Community.js
robbiecarlton: agreed

PR #204, opened 2017-01-06
GraphQL mutation

levity:


levity:
No need for all of you to review this -- one's enough, whoever can get to it soonest.

Connoropolous:
reviewing

api/graphql/schema.graphql
```diff
@@ -49,3 +53,16 @@ type Query {
...
   community(id: ID, slug: String): Community
 }
+
+input MeInput {
+  name: String
+  email: String
+  avatarUrl: String
+  bannerUrl: String
+  bio: String
+  url: String
```
Connoropolous: what led to the selection of these specific attributes, from the whole 'whitelist' that is under validateAndSave, and the difference between the graphql mutation and the possible mutation available via the original controller method

levity: This was the set that I was pretty confident we would continue to directly update as simple strings. I think we might want to create separate mutations for tags, password, and social media links. And then there are a few other fields that are not used anymore.

levity:
Here's an example of what a mutation looks like. You specify what you want to be returned just as you do for a query:

```graphql
mutation {
  updateMe(changes: {bio: "I am the very model"}) {
    id
    name
    email
    bio
    posts(first: 3) {
      title
    }
  }
}
```

The arguments can also be separated out:

```graphql
mutation ($changes: MeInput) {
  updateMe(changes: $changes) {
    id
    name
    email
    bio
    posts(first: 3) {
      title
    }
  }
}

# this part would go in the "Query Variables" section of GraphiQL

{
  "changes": {
    "bio": "I am the very model"
  }
}
```

api/graphql/schema.graphql
Connoropolous: makes sense to me 👍

PR #208, opened 2017-01-11
Send parent_post_id as part of commentAdded event payload

robbiecarlton:


robbiecarlton:
Depends on https://github.com/Hylozoic/hylo-redux/pull/234

levity:
some part of me wonders if this would be a good place to start switching to camelcase

PR #209, opened 2017-01-13
Project tagging breakup

lorenjohnson:
- [x] need to format children of a post, in the case of e.g. /noo/post/43?children=true. 
we currently just send the entire object as seen in the database back


Connoropolous:
@levity if I could get your eyes on this (the data migration) in the near future, before the rest of the review that'd be great :) 
https://github.com/Hylozoic/hylo-node/pull/209/commits/75fd7fb9d446a27bc3c4268b91acce901e449890

lib/migration.js
```diff
@@ -104,3 +104,59 @@ export const convertAllProjectsToPosts = () => {
...
+    Promise.resolve())
+  )
+}
+
+const findOfficialProjectTags = (projects) => {
+  return reduce(
+    projects,
+    (promise, project) => promise
+      .then(() => PostTag.query(qb => {
+          qb.where('post_id', '=', project.id)
```
levity: you don't need to specify `=`; if you omit it, it's assumed that's what you mean

api/controllers/PostController.js
```diff
@@ -102,6 +102,13 @@ const queryForFollowed = function (req, res) {
   }))
 }
 
+const queryForPost = function (req, res) {
```
levity: Since this is used for a child post I would name this `queryForChildPost`.

api/models/Tag.js
```diff
@@ -279,7 +279,7 @@ module.exports = bookshelf.Model.extend({
   },
 
   nonexistent: (names, communityIds) => {
-    if (names.length === 0 || communityIds.length === 0) return Promise.resolve({})
+    if (names.length === 0 || !communityIds || communityIds.length === 0) return Promise.resolve({})
```
levity: using lodash's `isEmpty` here would cover both the null and empty cases

Connoropolous: k, will switch!

Connoropolous: oh, actually, the two cases are opposite, it needs both separate

Connoropolous: hm, sorry again. I COULD switch it but I think it would make it less clear. So I'm still going to leave it

Connoropolous:
bing! it's ready

PR #210, opened 2017-01-23
Project activity cards

robbiecarlton:


api/controllers/PostController.js
```diff
@@ -54,6 +54,12 @@ const fetchAndPresentPosts = function (query, userId, relationsOpts) {
     data.posts.forEach((post, i) => normalizePost(post, buckets, i === data.posts.length - 1))
     return Object.assign(data, buckets)
   })
+  .then(data =>
```
levity: if you do this with `tap` then it's a bit briefer -- you can skip `return data` at the end

api/presenters/PostPresenter.js
```diff
@@ -138,9 +139,28 @@ const postListRelations = (userId, opts = {}) => {
...
+    q.orderBy('updated_at', 'desc')
+    q.limit(1)
+  })
+  .fetch({withRelated: postListRelations(userId, relationsOpts || {})})
+  .then(child => {
+    if (!child || Math.abs(post.updated_at.getTime() - child.get('updated_at').getTime()) > 1000) return post
+    child = postAttributes(child, userId, relationsOpts)
+    child.project = post
+    child.type = 'project-activity'
+    normalizePost(child, data, true)
```
levity: the third argument should only be set to `true` at the end of the whole iteration -- so you could either use the index from `map`, e.g.:

```js
Promise.map(data.posts, (p, i) => presentProjectActivity(post, data, userId, relationsOpts, i == data.posts.length - 1))
```

or you could just always omit the third argument here, and just call `uniqize` in `then` after Promise.map, like we do in SearchController.

PR #216, opened 2017-02-06
Tags for comments are also attached to the parent post

robbiecarlton:


test/unit/models/Tag.test.js
```diff
@@ -270,25 +270,49 @@ describe('Tag', () => {
...
+
+    it('preserves a tag from a comment, and removes a tag not in any of the comments', () => {
+      const post = factories.post({description: '<p>#preexisting</p>'})
+      const tag1 = factories.tag({name: 'preexisting'})
+      const tag2 = factories.tag({name: 'commenttag'})
+      const tag3 = factories.tag({name: 'unexpected'})
+      const comment = factories.comment()
+      return Promise.join(post.save(), tag1.save(), tag2.save(), tag3.save(), comment.save())
+      .then(() => Promise.join(
+        post.tags().attach(tag1),
```
levity: you can pass multiple objects into `attach` by passing an array as the first argument

test/unit/controllers/CommentController.test.js
```diff
@@ -279,7 +279,7 @@ describe('CommentController', function () {
       }))
     })
 
-    it('saves a new tag with description to the community', () => {
+    it('saves a new tag with description to the community and adds it to the parent post', () => {
```
levity: the test name changed but the test itself didn't

api/models/Post.js
```diff
@@ -90,6 +90,18 @@ module.exports = bookshelf.Model.extend({
     return this.belongsTo(LinkPreview)
   },
 
+  childTags: function (opts) {
```
levity: let's call this `getChildTags` since it's not a formal Bookshelf relation

api/models/Post.js
```diff
@@ -90,6 +90,18 @@ module.exports = bookshelf.Model.extend({
...
+    // this is part of the 'taggable' interface, shared with Comment
+    return this.load(['comments', 'comments.tags'], opts)
```
levity: you don't need to specify `comments` explicitly here, only `comments.tags`

levity: and then you could do `uniqBy('id', flatten(p.relations.comments.map(c => c.relations.tags.models)))`

api/models/Post.js
levity: actually `getTagsInComments` is more to the point

api/models/Tag.js
```diff
@@ -80,28 +80,32 @@ const updateForTaggable = ({ taggable, text, selectedTagName, tagDescriptions, u
...
-      name: t.get('name'),
-      selected: t.pivot.get('selected')
-    }))
-    const toAdd = uniqBy(lowerName, tagDifference(newTags, oldTags))
-    const toRemove = tagDifference(oldTags, newTags)
-    return Promise.all(flatten([
-      toRemove.map(tag => removeFromTaggable(taggable, tag, {transacting})),
-      toAdd.map(tag => addToTaggable(taggable, tag.name, tag.selected, tagDescriptions || {}, userId, {transacting}))
-    ]))
+    return taggable.load('tags')
```
levity: lost the `{transacting}` argument here

robbiecarlton:
@levity - made those changes. Good to merge?

PR #217, opened 2017-02-08
Preserve selected tag in posts when updating tags

robbiecarlton:


robbiecarlton:
Feel free to merge if it looks good to you

PR #218, opened 2017-02-14
Add last name to autocomplete search

lorenjohnson:
Turns out the existing PG full text search term search is configured to do exactly this, so I went ahead and used just that.

### DIFF
--- a/api/services/Search.js
+++ b/api/services/Search.js
@@ -155,7 +155,9 @@ module.exports = {
       }
 
       if (opts.autocomplete) {
-        qb.whereRaw('users.name ilike ?', opts.autocomplete + '%')
+        Search.addTermToQueryBuilder(opts.autocomplete, qb, {
+          columns: ['users.name']
+        })
       }
 
       if (opts.term) {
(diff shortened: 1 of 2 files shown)


api/services/Search.js
```diff
@@ -155,7 +155,9 @@ module.exports = {
       }
 
       if (opts.autocomplete) {
-        qb.whereRaw('users.name ilike ?', opts.autocomplete + '%')
+        Search.addTermToQueryBuilder(opts.autocomplete, qb, {
+          columns: ['users.name']
+        })
       }
```
lorenjohnson: One alternative which also works fine (though just slightly less performant) is:

```
qb.where(function () {
  this.orWhereRaw("split_part(users.name, ' ', 1) ILIKE ?", opts.autocomplete + '%')
  this.orWhereRaw("split_part(users.name, ' ', 2) ILIKE ?", opts.autocomplete + '%')
  this.orWhereRaw("split_part(users.name, ' ', 3) ILIKE ?", opts.autocomplete + '%')
})

```

PR #219, opened 2017-02-16
add total counts for pagination to our graphql api

levity:
So I was looking into the pagination bug for direct message threads, and started considering rewriting that code to use GraphQL, which uncovered the fact that I hadn't yet implemented a key part of pagination: getting the total number of results for a given query, so the client knows whether it's at the end of the result set yet.

The total number of results is not straightforward to add to the API as it currently is, because it's a property of the whole set of results, not any one instance.

The way this is handled by Relay is to add another level of indirection: instead of

```graphql
me {
  memberships(first: 3) {
    community {
      name
    }
  }
}
```

you might have

```graphql
me {
  memberships(first: 3) {
    edges {
      node {
        name
    }
    pageInfo {
      hasNextPage
    }
  }
}
```
(see the [spec docs](http://facebook.github.io/relay/graphql/connections.htm))

I wasn't prepared to go in that direction, so what I've done in this PR instead is:

```graphql
me {
  memberships(first: 3) {
    community {
      name
    }
  }
  membershipsTotal
}
```
where `membershipsTotal` is linked to whatever query parameters you used in `memberships` and does not run the query twice. Doesn't add any additional nesting and uses an easy-to-understand naming convention.

Opinions welcome.

### DIFF
--- a/api/graphql/fetcher.js
+++ b/api/graphql/fetcher.js
@@ -59,7 +59,6 @@ export default class Fetcher {
 
     const formatted = Object.assign(
       transform(attributes, (result, attr) => {
-        console.log(`${tableName} ${instance.id} ${attr} => ${camelCase(attr)} => ${instance.get(attr)}`)
         result[camelCase(attr)] = instance.get(attr)
       }, {}),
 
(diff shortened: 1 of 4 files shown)


Connoropolous:
aha! I was trying to use graphql to fetch communities and it wasn't working, but I see you fixed it in this branch! I had already looked this over once, but it went a bit over my head at the time, but I will have some thoughts on it very soon no doubt, as I'm using it now

levity:
Ha, I totally forgot about this PR.

PR #225, opened 2017-03-13
use an env var to allow access to axolotls profile

Connoropolous:
I think there will also be a tiny front-end tweak, to make it so that the 'message' bubble doesn't appear for axolotl on its profile... 

we'll save that for when it becomes an intelligent AI super-axolotl


Connoropolous:
on Hylo.com Axolotls profile id, for the AXOLOTL_ID env var is `13986`

PR #233, opened 2017-03-21
Oauth token json response, for FB and other mobile sign ins

Connoropolous:


api/controllers/SessionController.js
```diff
@@ -83,6 +83,14 @@ const finishOAuth = function (strategy, req, res, next) {
   return new Promise((resolve, reject) => {
     var respond = error => {
       if (error && error.stack) rollbar.handleError(error, req)
+      if (req.headers.accept === 'application/json') {
+        if (error) {
+          res.status(500).send(error)
```
levity: use res.serverError here -- it will skip the stack trace and also notify rollbar in production.

PR #234, opened 2017-03-29
Add comments to graphql

richchurcher:
Merging without tests to enable other branches to be created! 

Stops creator in graphql responses for posts and comments from being null.


api/graphql/schema.graphql
```diff
@@ -74,6 +74,20 @@ type Post {
...
+}
+
+type Comment {
+  id: ID
+  text: String
+  creator: Person
+  createdAt: String
+  createdFrom: String
+  active: Boolean
+  deactivatedReason: String
```
levity: we don't need these three `deactivated*` fields in the schema

PR #242, opened 2017-04-08
allow different GraphQL schema types to point to the same table

levity:
This PR extends the graphql-bookshelf bridge to allow multiple different GraphQL types to be backed by the same Bookshelf models.

It also uses this to add MessageThread and Message to the GraphQL schema, which are backed by Bookshelf models Post and Comment respectively.

The syntax for model specs (the output of `makeModels`) changes to support this in a few ways, all of which can be seen in the diffs. Overall, the purpose of these changes is for disambiguation: if multiple GraphQL types point to the same Bookshelf model, at the time when we have a Bookshelf model and need to map it back to a GraphQL type, we need to know which one to map it back to.

1. The top-level keys are now GraphQL schema type names, not database table names.

2. The syntax for aliases in `relations` changes:
    ```js
    // old
    {
      relations: [
        {creator: 'user'}
      ]
    }
    // new
    {
      relations: [
        {user: {alias: 'creator'}}
      ]
    }    
    ```

3. The change above was made in order to accommodate a new `typename` parameter in `relations`, which specifies the type name associated with a relation:
    ```js
    {
      relations: [
        {comments: {alias: 'messages', typename: 'Message'}}
      ]
    }
    ```

4. Since many different types have relations to e.g. `Post`, I added a shorthand for declaring that a GraphQL type should be used as the default for a bookshelf model, rather than repeatedly specify the same `typename`. This is the `isDefaultTypeForTable: true` property that was added to `Post`, `Person`, and `Comment`.

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -2,7 +2,8 @@
 // should be exposed through GraphQL, and what query filters should be applied
 // based on the current user's access rights.
 //
-// keys here are table names (except for "me")
+// keys in the returned object are GraphQL schema type names
+//
 export default function makeModels (userId, isAdmin) {
   // TODO: cache this?
   const myCommunityIds = () =>
@@ -13,8 +14,7 @@ export default function makeModels (userId, isAdmin) {
     isAdmin ? relation : relation.query(queryFn)
 
   return {
-    me: { // the root of the graph
-      typename: 'Me',
+    Me: { // the root of the graph
       model: User,
       attributes: [
         'id',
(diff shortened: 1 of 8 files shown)


api/graphql/makeModels.js
```diff
@@ -125,6 +139,21 @@ export default function makeModels (userId, isAdmin) {
         url: c => c.get('url'),
         imageUrl: c => c.get('image_url')
       }
+    },
+
+    MessageThread: {
+      model: Post,
+      attributes: ['id', 'created_at', 'updated_at'],
+      relations: [
+        'followers',
```
Connoropolous: We could leave this as 'followers' but in my front end work I was inclining towards calling them 'participants' which lends a little more clarity in a 'messaging' context. This seems like an opportunity to do that here too if we wanted...

richchurcher: Yeah that's a nice semantic nod to the overall community concept, I like it.

PR #243, opened 2017-04-08
More Person details in GraphQL response

richchurcher:
Adds location, bio, twitter_name, facebook_url, linkedin_url, url, location as possible GraphQL responses. Added to enable MemberProfile page.

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -54,7 +54,7 @@ export default function makeModels (userId, isAdmin) {
         'url',
         'location'
       ],
-      relations: ['communities', 'posts'],
+      relations: ['memberships', 'posts'],
       filter: nonAdminFilter(q => {
         q.where('users.id', 'in', Membership.query().select('user_id')
           .where('community_id', 'in', myCommunityIds()))
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -37,6 +37,9 @@ export default function makeModels (userId, isAdmin) {
       model: Membership,
       attributes: ['created_at', 'role', 'last_viewed_at'],
       relations: ['community']
+      filter: nonAdminFilter(q => {
+        q.where('communities_users.community_id', 'in', myCommunityIds())
+      })
     },
 
     users: {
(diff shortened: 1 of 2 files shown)


api/graphql/makeModels.js
```diff
@@ -42,8 +42,19 @@ export default function makeModels (userId, isAdmin) {
...
+        'avatar_url',
+        'banner_url',
+        'bio',
+        'twitter_name',
+        'linkedin_url',
+        'facebook_url',
+        'url',
+        'location'
+      ],
+      relations: ['memberships', 'posts'],
```
levity: If we do `memberships` instead of `communities`, we have to add a filter for Membership, basically:

```js
filter: nonAdminFilter(q => {
  q.where('communities_users.community_id', 'in', myCommunityIds())
})
```

Otherwise we'd be exposing memberships to communities that the viewer is not allowed to know about.

richchurcher: Good spotting! Thanks, I made that switch in too much of a hurry.

richchurcher: Is this the correct filter?
```javascript
      filter: nonAdminFilter(q => {
        q.where('users.id', 'in', Membership.query().select('user_id')
          .where('communities_users.community_id', 'in', myCommunityIds()))
      })
```
Or does it need to be:
```javascript
      filter: nonAdminFilter(q => {
        q.where('users.id', 'in', Membership.query().select('user_id')
          .where('community_id', 'in', myCommunityIds()))
          .where('communities_users.community_id', 'in', myCommunityIds()))
```
The latter, I think?

richchurcher: Actually that's not going to work either, GraphQL is erroring out... don't merge this yet :)

levity: You don't want to add the filter to `users`, but to `communities_users`. It's access to the latter that we want to control.

Btw, n.b., I have changes to the same files in #242. Feel free to review that one and I'll merge it.

richchurcher: Done!

api/graphql/makeModels.js
```diff
@@ -58,6 +58,7 @@ export default function makeModels (userId, isAdmin) {
       filter: nonAdminFilter(q => {
         q.where('users.id', 'in', Membership.query().select('user_id')
           .where('community_id', 'in', myCommunityIds()))
+          .where('communities_users.community_id', 'in', myCommunityIds())
```
levity: What I meant is that you want this line to go in a new `filter` prop in the `communities_users` section of the models file, not the `users` section. (In the new arrangement as of #242, that's the Membership section, not the Person section.) What we're trying to do is filter the `communities_users` rows that will be returned for a given user, not filter users based on their community membership. (We're already doing that in the line above, so putting this here is redundant.)

PR #244, opened 2017-04-10
find or create thread graphql + refactor

Connoropolous:
tests pass, and I also verified that it works as expected from hylo-redux still, the start a thread functionality

a query looks like: 
```
mutation {
  findOrCreateThread(data: {participantIds: ["2"]}) {
    id
    participants {
      id
      name
    }
  }
}
```

### DIFF
--- a/api/models/post/util.js
+++ b/api/models/post/util.js
@@ -31,6 +31,13 @@ export function validatePostCreateData (userId, data) {
   .then(ok => ok ? Promise.resolve() : Promise.reject(new Error('unable to post to all those communities')))
 }
 
+export function validateThreadData (data) {
+  if (!(data.participantIds && data.participantIds.length)) {
+    throw new Error('participipantIds can\'t be empty')
+  }
+  return Promise.resolve()
+}
+
 export const setupNewPostAttrs = function (userId, params) {
   const attrs = merge(Post.newPostAttrs(), {
     name: sanitize(params.name),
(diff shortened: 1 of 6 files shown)


api/models/post/util.js
```diff
@@ -192,14 +207,22 @@ export const createPost = (userId, params) =>
       {children: params.requests, transacting: trx}
     )))))
 
-export const createThread = (userId, params) =>
-  setupNewPostAttrs(userId, params)
+export const findOrCreateThread = (userId, participantIds) =>
+  Post.query(q => {
+    q.join('follows', 'follows.post_id', 'posts.id')
```
levity: i don't think you need to join here, because you don't reference `follows` explicitly anywhere else -- the `Follow.query()` lines create subqueries

api/models/post/util.js
```diff
@@ -192,14 +207,22 @@ export const createPost = (userId, params) =>
       {children: params.requests, transacting: trx}
     )))))
 
-export const createThread = (userId, params) =>
-  setupNewPostAttrs(userId, params)
+export const findOrCreateThread = (userId, participantIds) =>
+  Post.query(q => {
```
levity: check `Membership.inSameCommunity([userId].concat(participantIds))` first

api/models/post/util.js
```diff
@@ -31,6 +31,13 @@ export function validatePostCreateData (userId, data) {
   .then(ok => ok ? Promise.resolve() : Promise.reject(new Error('unable to post to all those communities')))
 }
 
+export function validateThreadData (data) {
+  if (!(data.participantIds && data.participantIds.length)) {
+    throw new Error('participantIds can\'t be empty')
```
levity: it's ok to use double quotes when there's a single quote in the string

Connoropolous:
@levity made the requested changes

levity:
:+1:

PR #246, opened 2017-04-12
Add an explicit moderator flag to gql

richchurcher:
Saves having to muck about with integer constants in the front end.


levity:
this means we can refactor some of our getters to be attributes instead too.

PR #249, opened 2017-04-13
Get parent post from comment

richchurcher:
Yet another addition! This one is to grab the post for a comment so we can display its name (and link to its id) in a comment card on the member profile overview.

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -144,7 +144,7 @@ export default function makeModels (userId, isAdmin) {
         'created_at'
       ],
       relations: [
-        {post: {alias: 'post'}},
+        'post',
         {user: {alias: 'creator'}}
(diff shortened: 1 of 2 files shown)


api/graphql/makeModels.js
```diff
@@ -144,6 +144,7 @@ export default function makeModels (userId, isAdmin) {
         'created_at'
       ],
       relations: [
+        {post: {alias: 'post'}},
```
levity: you can just have this be the string `'post'` if you don't need an alias

richchurcher: I wasn't sure if it would work in the singular! All the other relations were 'tablename-s' :)

richchurcher: Oh look at that, it does :)

PR #254, opened 2017-04-17
Add votes to graphql

richchurcher:
Required for the 'Upvotes' tab on MemberProfile. Please check I have the right nonadminFilter? :)

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -62,7 +62,12 @@ export default function makeModels (userId, isAdmin) {
         'url',
         'location'
       ],
-      relations: ['comments', 'memberships', 'posts'],
+      relations: [
+        'comments',
+        'memberships',
+        'posts',
+        'votes'
+      ],
       filter: nonAdminFilter(q => {
         q.where('users.id', 'in', Membership.query().select('user_id')
           .where('community_id', 'in', myCommunityIds()))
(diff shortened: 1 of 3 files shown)


api/graphql/makeModels.js
```diff
@@ -202,6 +207,23 @@ export default function makeModels (userId, isAdmin) {
...
+      ],
+      relations: [
+        'post',
+        { user: { alias: 'voter' } }
+      ],
+      filter: nonAdminFilter(q => {
+        q.where('votes.post_id', 'in', PostMembership.query().select('post_id')
+          .where('community_id', 'in', myCommunityIds()))
+      }),
+      isDefaultTypeForTable: true
```
levity: you should be able to omit this because there isn't a different GraphQL type that also uses the Vote model

api/graphql/makeModels.js
```diff
@@ -213,7 +213,7 @@ export default function makeModels (userId, isAdmin) {
       model: Vote,
       attributes: [
         'id',
-        'created_at'
+        'date_voted'
```
levity: Let's do a `createdAt` getter that relies on `date_voted` so that the API's outer surface is consistent

richchurcher: Yeah I like that better :)

PR #256, opened 2017-04-19
Settings gql

robbiecarlton:


api/graphql/mutations.js
```diff
@@ -26,6 +26,16 @@ function convertGraphqlCreateData (data) {
   }, data))
 }
 
+function convertGraphqlUserSettingsData (data) {
```
Connoropolous: might be worthwhile to convert this one to behaving as a promise to match the one above, or converting the one above. feels awkward to have the two behave opposingly

robbiecarlton: that makes sense. I'm a little resistant to doing that as I think this code is already awkward and that there is a larger refactor that is close to due here. So it's like: let's not put lipstick on the pig.
But maybe this is how we'll be doing it for awhile, in which case makes sense to standardize the pattern as you suggest.

api/graphql/mutations.js
```diff
@@ -10,9 +10,20 @@ import {
   validateThreadData
 } from '../models/post/util'
 
+function convertGraphqlUserSettingsData (data) {
+  return transform(data, (result, value, key) => {
```
levity: just fyi, if you're not going to remove/add/merge keys, `mapValues` does the job too

PR #258, opened 2017-04-23
fetchMany in graphql schema

levity:
sketch out a way of adding root endpoints that return query sets to the
graphql api, while delegating their implementation into makeModels

all existing tests pass, but this hasn't been itself tested -- i'm just offering it for discussion

### DIFF
--- a/api/graphql/searchQuerySet.js
+++ b/api/graphql/searchQuerySet.js
@@ -60,6 +60,6 @@ const shims = {
   },
 
   forUsers: options => {
-    // TODO
+    return options
   }
 }
--- a/api/graphql/searchQuerySet.js
+++ b/api/graphql/searchQuerySet.js
@@ -1,4 +1,5 @@
 import { mapValues } from 'lodash'
+import { isNull, isUndefined, omitBy } from 'lodash/fp'
 import { PAGINATION_TOTAL_COLUMN_NAME } from '../../lib/graphql-bookshelf-bridge/util/applyPagination'
 
 export default function searchQuerySet (searchName, options) {
@@ -24,7 +25,7 @@ export function sanitizeOptions (name, options) {
   if (!shim) throw new Error(`no option shim for ${name}`)
 
   const withDefaults = Object.assign({}, defaultOptions.all,
-    defaultOptions[name], options)
+    defaultOptions[name], omitBy(x => isNull(x) || isUndefined(x), options))
   return shim(withDefaults)
 }
 
--- a/api/graphql/searchQuerySet.js
+++ b/api/graphql/searchQuerySet.js
@@ -1,26 +1,48 @@
 import { mapValues } from 'lodash'
+import { PAGINATION_TOTAL_COLUMN_NAME } from '../../lib/graphql-bookshelf-bridge/util/applyPagination'
 
 export default function searchQuerySet (searchName, options) {
-  const opts = shimOptions(searchName, options)
-  return Search[searchName](opts).fetchAll()
-  .then(({ length, models }) => {
-    const items = models
+  return Search[searchName](sanitizeOptions(searchName, options))
+}
 
+export function fetchSearchQuerySet (searchName, options) {
+  return searchQuerySet(searchName, options).fetchAll()
+  .then(({ length, models }) => {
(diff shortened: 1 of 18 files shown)


levity:
@richchurcher and @lorenjohnson; here's what i came up with re: our recent discussion in Slack. what i like about this is that the code in `lib/graphql-bookshelf-bridge` doesn't have to know anything about argument names or the SQL query clauses they correspond to; it just passes them all through to `makeModels`.

richchurcher:
Cool! I'll see if I can adapt it for the people thing and report back...

...and then it can be...

...AUTOMATIC FOR THE PEOPLE... :notes:

levity:
oh, right, you were working on people. how did that escape my attention? apparently i'm drunk on hot cocoa. yes, it should work fine. See `Search.forUsers`

lorenjohnson:
LGTM--I think, seems pretty much the way to do it. I will grok this more once I try and apply it for People in the Editor mentions, unless @richchurcher gets there first (we will talk). Of course I'm also require pulling #topics so I need to look into how to get to that data and put it in this pattern as well.

levity:
aaand now I got to the point where I need this too, for All Communities. Will keep working off this branch.

api/graphql/searchQuerySet.js
```diff
@@ -1,26 +1,48 @@
 import { mapValues } from 'lodash'
+import { PAGINATION_TOTAL_COLUMN_NAME } from '../../lib/graphql-bookshelf-bridge/util/applyPagination'
 
 export default function searchQuerySet (searchName, options) {
```
levity: this splits into `searchQuerySet` and `fetchSearchQuerySet`; the former just prepares a Bookshelf query, and the latter actually runs it. There's definitely some DRYing up that can happen around this, since the adding of `hasMore` to the return value happens both below and in `fetcher.js`.

api/services/Search/forPosts.js
```diff
@@ -9,7 +10,7 @@ export default function forPosts (opts) {
 
     // this counts total rows matching the criteria, disregarding limit,
     // which is useful for pagination
-    qb.select(bookshelf.knex.raw('posts.*, count(*) over () as total'))
+    countTotal(qb, 'posts', opts.totalColumnName)
```
levity: it was necessary to parameterize the name that `Search` uses for the total column; in the new GraphQL API code it should be the value of `PAGINATION_TOTAL_COLUMN_NAME`, but we have to leave the old column name as default so we don't break the old site.

lib/graphql-bookshelf-bridge/util/initDataLoaders.js
```diff
@@ -23,7 +23,9 @@ export default function initDataLoaders (spec) {
   // general-purpose query cache, for relational SQL queries that aren't just
   // fetching objects by ID.
   loaders.queries = new DataLoader(
-    queries => Promise.map(queries, q => q.fetch()),
+    queries => Promise.map(queries, ({ query, method }) => {
+      return method ? query[method]() : query.fetch()
```
levity: this here works around something in Bookshelf's API that I have always thought of as a wart: when you have a relation, `fetch` gets the whole set and you use `fetchOne` to get a single row, but when you have a collection built from `Post.query(q => { ... })`, `fetch` gets a single result and `fetchAll` gets the whole set. We were actually relying on the ambiguity of `fetch` here to handle both `fetchRelation` and `fetchOne`, but now that we're using `Search`, we can't do so anymore.

lib/graphql-bookshelf-bridge/fetcher.js
```diff
@@ -21,42 +21,38 @@ export default class Fetcher {
...
       this.loaders[typename].prime(instance.id, instance)
       return instance
     })
   }
 
+  fetchMany (typename, args) {
+    const { fetchMany } = this._getModel(typename)
+    let query = fetchMany(args)
+    return this._loadMany(query, {
+      method: 'fetchAll',
```
levity: see my comment on `initDataLoaders` for the explanation of this

levity:
just looking for a quick once-over from one of you before i merge

PR #259, opened 2017-04-26
Add description and moderators to community gql

robbiecarlton:


api/graphql/makeModels.js
```diff
@@ -121,12 +121,16 @@ export default function makeModels (userId, isAdmin) {
...
         'slug',
+        'description',
         'created_at',
         'avatar_url',
         'banner_url',
         'memberCount',
         'postCount'
       ],
+      relations: [
+        'moderators'
```
levity: let's only add querySet-style relations from here on out

PR #260, opened 2017-04-27
Add "vote" mutation to GraphQL

robbiecarlton:


levity:
I think it'd be much more resilient to latency, as well as a bit simpler in code, to pass along the desired vote value to the API (e.g. true or false, 1 or -1). For example, if the value is always set from the frontend, you could imagine clicking the button many times, watching the UI flip back and forth, and then being absolutely certain that the state you get on the backend eventually 

Does that make sense? Sorry to bring this up after you've already implemented it as a toggle.

robbiecarlton:
I switched to a model that sends a 'didVote' flag, and added tests

levity:
maybe `isUpvote` instead? `didVote` is confusing to me; evokes `componentDidMount` and the like

PR #263, opened 2017-05-08
add filtering by topic to graphql post search

Connoropolous:
topic must be specified by its ID


levity:
Requiring an ID would be a good thing to enforce in sanitizeOptions and add a test for in searchQuerySet.test.js

Connoropolous:
@levity like so?

PR #264, opened 2017-05-08
add community topics

Connoropolous:
so this PR makes it so that if you're interested in the TOTAL number of posts for a topic, look at the Topic itself, and grab `postsTotal` off it, but if you're interested in the total number of posts for a topic within a community, grab it off of the CommunityTopic using `postsTotal`

### DIFF
--- a/api/models/Tag.js
+++ b/api/models/Tag.js
@@ -294,6 +294,24 @@ module.exports = bookshelf.Model.extend({
     })
   },
 
+  taggedPostCount: tagId => {
+    return bookshelf.knex('posts_tags')
+    .where({tag_id: tagId})
+    .count()
+    .then(rows => Number(rows[0].count))
+  },
+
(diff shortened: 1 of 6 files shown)


levity:
LGTM

api/models/Tag.js
```diff
@@ -294,6 +294,24 @@ module.exports = bookshelf.Model.extend({
     })
   },
 
+  taggedPostCount: tagId => {
+    return bookshelf.knex('posts_tags')
```
Connoropolous: @levity should we be expanding this query, and another similar one in CommunityTag to filter for only `active` posts?

levity: yes

Connoropolous: K, will add that

Connoropolous:
@levity k, fixed that, and added tests

PR #268, opened 2017-05-09
Equalises knex_migrations with directory

richchurcher:
Obtain a directory listing of `./migrations`, and remove everything in `knex_migrations` which isn't one of those files. To run: `npm run update-migrations`.

This is to address the situation where an existing developer has a couple of hundred migrations in their database, and we have removed all but the 2017 migrations from the directory. Knex will complain if any files are missing from `knex_migrations`.


richchurcher:
YW!

PR #269, opened 2017-05-09
Connections (WIP)

richchurcher:
Store data on recent connections in the backend for use on the new messages contacts display. Could potentially also be used as a friends list, mentions list, etc. Currently just supports 'messages' type.

I've left the question of how to maintain this table (remove old entries) open, and just assume for the moment that everything in it is relevant. Per discussion in Slack, we'll need to have some kind of regular job to prune it.

Allows the following query:
```
{
  connections (first: 10) {
    items {
      id
      person {
        id
        name
        avatarUrl
      }
      type
      createdAt
      updatedAt
    }
  }
}
```
and mutation:
```
mutation MakePersonConnections ($data: PersonConnectionInput) {
  createOrUpdatePersonConnections (data: $data) {
    id
    person {
      id
      name
      avatarUrl
    }
    type
    createdAt
    updatedAt
  }
}
```


api/models/index.js
```diff
@@ -11,6 +11,7 @@ export const init = () => {
   global.Promise = Promise
 
   global.bookshelf = Bookshelf(Knex(knexfile[process.env.NODE_ENV]))
+  global.bookshelf.plugin('bookshelf-returning')
```
richchurcher: @levity just wanted to flag this for your attention, in case it has side-effects with other things we're doing that I don't know about! I just wanted to avoid fetch-after-create and fetch-after-update hits on the db, by using `{ returning: '*' }`. Potentially we could save quite a few hits from other models too by doing this.

levity:
This looks good overall but I have some concern about the names. I think `UserConnection` would be clearer since `Connection` by itself means many different things. And the relation could be called `otherUser` or something like that. Better to have it be a noun.

richchurcher:
Yeah I definitely wasn't sure about `with`, I was kinda trying it on for size :) The only issue I have with `UserConnection` is that it breaks our naming scheme on the front end, so... would you have the model be `UserConnection`, and then translate to `PersonConnection` or `MemberConnection` at the GraphQL layer?

levity:
Yes, PersonConnection.

richchurcher:
Thanks! Um, I'm not sure I'm the best person to ask because I have so little experience with ORMs... unless you count EntityFramework in C#! I think everybody's enjoying the process of being able to ignore the old API and use the GraphQL. I do wonder how it's going to scale... but I think that's a not terribly scientific observation based on the number of layers between the GraphQL request and the database. I sometimes find myself wanting to just issue a knex query and be done with it :smiling_imp:

PR #270, opened 2017-05-11
current user in post comments

levity:


api/models/Post.js
```diff
@@ -9,10 +9,17 @@ import { fulfillRequest, unfulfillRequest } from './post/request'
 import EnsureLoad from './mixins/EnsureLoad'
 import { countTotal } from '../../lib/util/knex'
 
-const commentersQuery = (limit, post) => q => {
+const commentersQuery = (limit, post, currentUserId) => q => {
   q.select('users.*', 'comments.user_id')
   q.join('comments', 'comments.user_id', 'users.id')
-  q.where('comments.post_id', '=', post.id)
+  if (currentUserId) {
```
levity: To order by a specific user ID:

```js
q.orderBy(bookshelf.knex.raw(`case when user_id = ${id} then -1 else user_id end`)
```

levity: @Connoropolous, to order by a specific user ID, do this:

```js
q.orderBy(bookshelf.knex.raw(`case when user_id = ${id} then -1 else user_id end`)
```

Connoropolous: ah, wow, ok cool

Connoropolous:
@levity ready for review

levity:
i can't approve it because i created it :) so i'll just merge

PR #272, opened 2017-05-12
Notifications graphQL

robbiecarlton:


api/graphql/notificationQuerySet.js
```diff
@@ -0,0 +1,37 @@
...
+    q.select(bookshelf.knex.raw('notifications.*, count(*) over () as total'))
+    q.orderBy('activities.created_at', 'desc')
+
+    Activity.filterInactiveContent(q)
+    if (community) Activity.joinWithCommunity(community.id, q)
+  })
+}
+
+export function fetchNotificationQuerySet (community, userId, options) {
+  return notificationQuerySet(community, userId, options).fetchAll({withRelated: [
```
robbiecarlton: Wasn't sure about putting all this in the fetchAll here. Am I missing something clever we can do in graphQL?

levity: Basically you shouldn't need to use a getter and an explicit query set, because you don't need to search notifications. It can be a relation. That will simplify some of this dramatically. But then you do lose the ability to use the syntatic sugar methods to skip having to go through `activity` explicitly to get to `post` or `comment`. I think that's probably for the best in order to avoid this gargantuan `fetchAll`.

robbiecarlton: so I  should include Activity as a separate model in graphQL and the redux orm?

levity: yep

robbiecarlton:
@levity - good to merge this?

api/graphql/notificationQuerySet.js
levity: this file can be removed now, no?

api/controllers/ActivityController.js
```diff
@@ -1,6 +1,6 @@
 const userColumns = q => q.column('users.id', 'name', 'avatar_url')
 
-const actionFromReason = reason => {
+export const actionFromReason = reason => {
```
levity: this is unnecessary i think

PR #273, opened 2017-05-15
Create UserConnections on thread message create

richchurcher:
Do a createOrUpdate on connections between the current user and participants in a message thread each time user creates a message.

Remove old mutation code, handle this in backend!


api/models/comment/createAndPresentComment.js
```diff
@@ -106,3 +107,10 @@ const notifySockets = (comment, post) => {
     return post.pushCommentToSockets(comment)
   }
 }
+
+const createOrUpdateConnections = (userId, existingFollowers) => comment => {
+  // Deliberately non-blocking (don't wait for promise to resolve/reject)
+  existingFollowers
+    .filter(f => f !== userId)
+    .forEach(follower => UserConnection.createOrUpdate(userId, follower, 'message'))
+}
```
richchurcher: @levity Is this ok? I didn't want to wait for the promise (hence not returning it to the `.tap`, which will only block if promise is returned) since there might be quite a few queries involved in a large thread. However, if I do this I should probably log errors somewhere with a `.catch`... how do we handle logging?

levity: We can't ever break the promise chain; it leads to unspecified behavior. If you don't want the response to block on a slow operation, look at the way we use Queue.classMethod elsewhere in the project to push work over to [kue](https://www.npmjs.com/package/kue).

richchurcher:
It does cause one of the tests to fail though, will check on that.

richchurcher:
Ok so @levity at your discretion I say go ahead and call this one ready for merge. I haven't given up on the testing suite but I don't want it to be a blocker anymore!

PR #279, opened 2017-05-20
graphql mutation and API endpoint for updating a post and related updatePost abstraction

lorenjohnson:


api/models/post/util.js
```diff
@@ -1,7 +1,6 @@
 import { difference, has, isEqual, pick, some } from 'lodash'
-import updateChildren from './updateChildren'
 
-const updateMedia = (post, type, url, remove, transacting) => {
+export function updateMedia (post, type, url, remove, transacting) {
```
levity: let's not `export` unnecessarily

PR #280, opened 2017-05-21
Make tags.id bigint

richchurcher:
This was a fiddly little thing. In order for it to work, I needed to provide the knex object used in migrations (which is normally set up to use a transaction for each migration I believe) to the view create/drop functions in FullTextSearch. They will use the old `bookshelf.knex` at other times.

Took me quite a long time to realise I couldn't do 
```
  .then(FullTextSearch(createView(null, knex)))
```
instead of 
```
  .then(() => FullTextSearch(createView(null, knex)))
```
Oh well, live and learn :grinning:

### DIFF
--- a/api/services/FullTextSearch.js
+++ b/api/services/FullTextSearch.js
@@ -4,13 +4,13 @@ const tableName = 'search_index'
 const columnName = 'document'
 const defaultLang = 'english'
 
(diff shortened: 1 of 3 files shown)


levity:
i went ahead and merged this myself and ran the migration in production.

PR #281, opened 2017-05-23
Update database seed

richchurcher:
Some tooling I've been using. Generates a fake database with:
- a user account and password
- lots of message threads pre-populated with people from the test community
- lots of posts
- lots of other communities (which the test user is not a member of, yet)
- lots of other users with avatars, profile background images, etc

I've guarded it pretty carefully to avoid accidental usage (it only runs when NODE_ENV is set correctly, and confirms prior to run). It'd just be handy to have it merged so I can stop switching branches when I need it!


seeds/dummy/dummy.js
```diff
@@ -0,0 +1,334 @@
+const bcrypt = require('bcrypt')
```
levity: this doesn't need to be `dummy/index`?

richchurcher: Nah, knex just shotguns every .js file in the specified directory :gun:

levity:
ah, i see. well in that case, :shipit:

PR #282, opened 2017-05-23
Add updateCommunitySettings to graphQL

robbiecarlton:


api/graphql/makeModels.js
```diff
@@ -147,7 +147,9 @@ export default function makeModels (userId, isAdmin) {
         'avatar_url',
         'banner_url',
         'memberCount',
-        'postCount'
+        'postCount',
+        'welcome_message',
```
levity: i don't think we actually need welcome_message for the new UI

api/graphql/mutations.js
```diff
@@ -118,3 +118,8 @@ export function unlinkAccount (userId, provider) {
   })
   .then(() => ({success: true}))
 }
+
+export function updateCommunitySettings (id, changes) {
+  return Community.find(id)
+  .then(community => community.update(convertGraphqlData(changes)))
```
levity: you need to validate that the user has permission to do this (i.e. is a moderator)

api/graphql/schema.graphql
```diff
@@ -351,6 +354,26 @@ input TopicSubscriptionInput {
...
 
+input CommunityInput {
+  name: String
+  description: String
+  location: String
+  welcomeMessage: String
+  bannerUrl: String
+  avatarUrl: String
+  settings: CommunitySettingsInput
+  leaderId: ID
```
levity: you can drop this as well as welcomeMessage

api/graphql/schema.graphql
```diff
@@ -351,6 +354,26 @@ input TopicSubscriptionInput {
...
+  betaAccessCode: String
```
levity: we don't need to expose this functionality yet (and in any case it'll be called something else)

PR #283, opened 2017-05-26
remove TopicSubscription from schema

levity:
put newPostCount and isSubscribed on CommunityTopic instead

also remove Search.forCommunityTopics and use the new relation
filtering option

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -152,7 +152,24 @@ export default function makeModels (userId, isAdmin) {
       ],
       relations: [
         {moderators: {querySet: true}},
-        {tagFollows: {querySet: true, alias: 'topicSubscriptions'}}
+        {communityTags: {
+          querySet: true,
+          alias: 'communityTopics',
+          filter: (relation, { autocomplete, subscribed }) => relation.query(q => {
+            if (autocomplete) {
+              q.join('tags', 'tags.id', 'communities_tags.tag_id')
+              q.whereRaw('tags.name ilike ?', autocomplete + '%')
+            }
+
+            if (subscribed) {
+              q.join('tag_follows', 'tag_follows.tag_id', 'communities_tags.tag_id')
+              q.where({
+                'tag_follows.community_id': relation.relatedData.parentId,
+                'tag_follows.user_id': userId
+              })
+            }
+          })
+        }}
       ],
       getters: {
         popularSkills: (c, { first }) => c.popularSkills(first),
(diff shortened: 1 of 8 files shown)


levity:
Required for Hylozoic/hylo-evo#101

api/graphql/mutations.js
```diff
@@ -83,18 +83,16 @@ export function updateMembership (userId, { id, data }) {
 
   return Membership.query().where({id, user_id: userId})
   .update(whitelist)
-  .returning('id')
-  .then(ids => Membership.where('id', ids[0]).fetch())
+  .then(() => ({success: true}))
```
Connoropolous: why go from returning the membership to returning success true? is the front-end prepared for this change?

As I read further I interpret this is because we're prepping mutations to mostly just return this GenericResult type?

api/graphql/makeModels.js
```diff
@@ -152,7 +152,24 @@ export default function makeModels (userId, isAdmin) {
       ],
       relations: [
         {moderators: {querySet: true}},
-        {tagFollows: {querySet: true, alias: 'topicSubscriptions'}}
+        {communityTags: {
+          querySet: true,
+          alias: 'communityTopics',
+          filter: (relation, { autocomplete, subscribed }) => relation.query(q => {
```
Connoropolous: just for the sake of learning and understanding, what's `relation` here?

api/graphql/mutations.js
```diff
@@ -83,18 +83,16 @@ export function updateMembership (userId, { id, data }) {
...
+export function updateCommunityTopic (userId, { id, data }) {
   const whitelist = mapKeys(pick(data, 'newPostCount'), (v, k) => snakeCase(k))
   if (isEmpty(whitelist)) return Promise.resolve(null)
 
-  return TagFollow.query().where({id, user_id: userId})
-  .update(whitelist)
-  .returning('id')
-  .then(ids => TagFollow.where('id', ids[0]).fetch())
+  return CommunityTag.where({id}).fetch()
+  .then(ct => ct.tagFollow(userId).query().update(whitelist))
```
Connoropolous: instead of letting the client actually SET newPostCount for their tagFollow, shouldn't we maybe call it `reset` and set the value in the backend? there's no reason a client should be able to control the specific value themselves, just clear it I think

api/graphql/mutations.js
levity: The front-end already ignores the response, so I figured there was no need to provide one. It's not a change for mutations in general, just ones where the front-end doesn't need any data about the outcome.

api/graphql/makeModels.js
levity: the return value of `someCommunity.communityTags()`

api/graphql/mutations.js
levity: that's fair -- i was generalizing a bit prematurely here, with the expectation that there would be other fields we'd want to update in the future

levity:
I'll work out the All Communities case in another PR. It's a bit non-trivial since we need to aggregate when a user is subscribed to the same topic in multiple communities.

PR #286, opened 2017-06-01
See all topics

robbiecarlton:
There are a couple of changes to the live database that will go with this one. Specifically in CommuntiyTags, setting updated_at and created_at when they're not currently set, and running consolidateFollowerCount on all Community Tags.

### DIFF
--- a/api/models/CommunityTag.js
+++ b/api/models/CommunityTag.js
@@ -1,3 +1,5 @@
+/* eslint-disable camelcase */
+
 module.exports = bookshelf.Model.extend({
   tableName: 'communities_tags',
 
@@ -41,10 +43,10 @@ module.exports = bookshelf.Model.extend({
   },
 
   consolidateFollowerCount: function () {
-    this.followerCount()
-    .then(followers => {
-      if (followers === this.get('followers')) return Promise.resolve()
-      return this.save({followers})
+    return this.followerCount()
+    .then(num_followers => {
+      if (num_followers === this.get('num_followers')) return Promise.resolve()
+      return this.save({num_followers})
     })
   }
 
--- a/api/models/CommunityTag.js
+++ b/api/models/CommunityTag.js
@@ -51,7 +51,7 @@ module.exports = bookshelf.Model.extend({
 }, {
 
   create (attrs, { transacting } = {}) {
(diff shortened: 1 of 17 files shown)


api/models/CommunityTag.js
```diff
@@ -38,12 +38,20 @@ module.exports = bookshelf.Model.extend({
 
   followerCount: function () {
     return Tag.followersCount(this.get('tag_id'), this.get('community_id'))
+  },
+
+  consolidateFollowerCount: function () {
+    this.followerCount()
```
levity: missing `return` here

migrations/20170531132953_add-followers-to-communties_tags.js
```diff
@@ -0,0 +1,11 @@
+exports.up = function (knex, Promise) {
+  return knex.schema.table('communities_tags', table => {
+    table.integer('followers').defaultTo(0)
```
levity: i feel like `num_followers` would be a bit clearer, like we have `num_votes` for `posts`

test/unit/models/TagFollow.test.js
```diff
@@ -0,0 +1,155 @@
...
+      .then(() => TagFollow.where(attrs).fetch())
+      .then(tagFollow => {
+        expect(tagFollow).not.to.exist
+      })
+    })
+  })
+
+  describe('#subscribe', () => {
+    it("creates a TagFollow when there isn't one only if isSubscribing", () => {
+      const attrs = {
```
levity: this assignment should be pulled out into a `beforeEach` because you use it several times

migrations/20170605161512_rename-followers.js
```diff
@@ -1,11 +1,13 @@
 exports.up = function (knex, Promise) {
   return knex.schema.table('communities_tags', table => {
-    table.integer('followers').defaultTo(0)
+    table.dropColumn('followers')
```
levity: pretty sure this migration will fail on production because `followers` doesn't exist

robbiecarlton: hmm. Yes. My database and migrations got a little messed up somewhere along the way I think. Removed the references to 'followers' now.

PR #288, opened 2017-06-06
Only return active posts from graphQL

robbiecarlton:


api/graphql/makeModels.js
```diff
@@ -126,7 +128,7 @@ export default function makeModels (userId, isAdmin) {
         'followers',
         'linkPreview'
       ],
-      filter: nonAdminFilter(sharedPostMembership('posts', userId)),
+      filter: nonAdminFilter(composeFilters(activePost, sharedPostMembership)('posts', userId)),
```
levity: The `tableName, userId, q` signature is pretty specific to the membership filters. It would make more sense to me to see something like:

```js
composeFilters(
  activePost,
  nonAdminFilter(sharedPostMembership('posts', userId))
)
```

robbiecarlton:
@levity - how's that?

levity:
Looks great--but now you can just use lodash.flow :)

robbiecarlton:
hah! Ok, done. I'm gonna merge.

PR #289, opened 2017-06-07
first time use of findOrCreateThread would break, and so would first comment

Connoropolous:
this would happen when using hylo-redux... opening the modal to start a new thread, then trying to send a comment. the user feedback for it not working wasn't good/implemented

related to the calling of createAndPresentComment (at comment creation)
screen shot 2017-06-06 at 9 06 54 pm

related to the calling of findOrCreateThread
screen shot 2017-06-06 at 9 07 04 pm

### DIFF
--- a/api/models/post/findOrCreateThread.test.js
+++ b/api/models/post/findOrCreateThread.test.js
@@ -1,4 +1,4 @@
-import { validateThreadData } from './findOrCreateThread'
+import { validateThreadData, createThread } from './findOrCreateThread'
 
 describe('validateThreadData', () => {
   var user, userSharingCommunity, userNotInCommunity, inCommunity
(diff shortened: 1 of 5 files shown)


api/models/post/findOrCreateThread.test.js
```diff
@@ -37,3 +37,29 @@ describe('validateThreadData', () => {
...
+    ).then(function () {
+      return Promise.join(
+        user.joinCommunity(community),
+        user2.joinCommunity(community)
+      )
+    })
+  })
+
+  it('creates and returns a new thread', () => {
+    createThread(user.id, [user2.id]).then(p => {
```
levity: missing `return` here

PR #290, opened 2017-06-08
Notifications socket updates

richchurcher:
This:
 - provides for socket.io updates from the worker process
 - updates the user socket with new notifications

Let me know if there's a better way to construct the object from Bookshelf, it feels a little klunky.

### DIFF
--- a/api/controllers/PostController.js
+++ b/api/controllers/PostController.js
@@ -422,11 +422,11 @@ const PostController = {
     .then(() => res.ok({}))
   },
 
-  subscribeToThreads: function (req, res) {
+  subscribeToUpdates: function (req, res) {
     joinRoom(req, res, 'user', req.session.userId)
   },
 
-  unsubscribeFromThreads: function (req, res) {
+  unsubscribeFromUpdates: function (req, res) {
     leaveRoom(req, res, 'user', req.session.userId)
   }
 }
--- a/api/models/Notification.js
+++ b/api/models/Notification.js
@@ -298,10 +298,19 @@ module.exports = bookshelf.Model.extend({
   updateUserSocketRoom: function (userId) {
     return getSailsInstance()
       .then(sails => {
+        const actor = Object.assign({},
+          this.actor().pick([ 'id', 'name' ]),
+          { avatarUrl: this.actor().get('avatar_url') }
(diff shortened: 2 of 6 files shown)


levity:
Does loading the `http` hook make sails try to start listening on a port? I remembered that workers already have a stripped-down sails instance (see worker.js and skiff.js), one in which I'd intentionally disabled the `http` and `sockets` hooks because it was trying to listen on the same port as the web process.

Perhaps `sails.lift` causes this but `sails.load` doesn't?

levity:
Although if the only purpose of this code is to get sockets working, I wonder if it wouldn't be clearer to have a `getSailsSockets` method that returns `sails.sockets`.

richchurcher:
@levity Yeah, I did think of having a sockets-only version but my thinking went along the lines of: "If we need a sails instance in the future for a worker we'll need to repeat this." Since it was just as easy to make it slightly more generic, I opted for that but don't feel strongly about it.

The `http` hook: it definitely listens, unfortunately. This author (three years ago) worked around by rewriting the pubsub hook to not depend on sockets (which depends on http): https://blog.thesparktree.com/reusing-sailsjs-waterline-models-in-background . 

Alternatively, there's actually no reason we couldn't just write raw socket.io code as far as I can see, and write to the rooms in redis. Let me know if you want me to explore that.

levity:
if the http hook is unable to grab the port it wants, does it still stay up and allow the websocket behavior we want?

levity:
welp, i just tested this locally and no harm came of it. i'm going to merge it myself to keep an eye on what happens in production.

levity:
Can you try this in production and verify that it works as expected?

richchurcher:
@levity Sure, sorry I didn't get to it sooner! I guess it doesn't throw if it can't bind the port to listen? The networking bit of the dynos docs suggests that workers can't get inbound network requests anyway, so that may be why we're getting away with it: https://devcenter.heroku.com/articles/dynos#networking

richchurcher:
Yep, it's working on prod:

notification_frames

I guess call it done for now, think about if we have to get all custom with it later?

levity:
Yep, sounds good!

PR #297, opened 2017-06-16
Notification settings

robbiecarlton:


api/graphql/schema.graphql
```diff
@@ -99,8 +106,16 @@ type Membership {
   lastViewedAt: String
   newPostCount: Int
   community: Community
+  settings: MembershipSettings
 }
 
+type MembershipSettings {
+  send_push_notifications: Boolean
```
robbiecarlton: I wasn't sure whether to change these or the user settings to camel case. I know we've done that elsewhere, but this would be adding a little extra complexity in a bunch of places and not sure what it buys us.

levity: Hmm. Basically what it would buy us is consistency: we'd be able to rely on the rule "All attributes in the API and frontend are camel case." It's pretty straightforward to use lodash.snakeCase to convert back inside the backend.

robbiecarlton:
@levity - I changed the settings to camelcase.

robbiecarlton:
@levity - made those changes and added tests. Good to merge?

PR #299, opened 2017-06-22
fixes issue where saving settings throws an error

razorman8669:
https://trello.com/c/ZWvyEYuK/463-evo-cannot-change-name-from-settings-page

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -120,7 +120,7 @@ module.exports = bookshelf.Model.extend(merge({
   setSanely: function (attrs) {
     const saneAttrs = omit(attrs, 'settings')
 
-    if (saneAttrs.twitter_name) {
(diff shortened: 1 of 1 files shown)


levity:
weird. what is it when it's defined but not a string?

razorman8669:
@levity it's an empty object:  `{ }`

levity:
Huh. @robbiecarlton any sense of if that's a frontend bug?

robbiecarlton:
@levity - that would make sense. I'll take a look.

levity:
Right so if it's a frontend bug then there'd be no need for this change :) so let's hold off on merging till that's clear

levity:
bump

robbiecarlton:
I'll take a look at this soon.

PR #300, opened 2017-06-26
graphql for LinkPreview

lorenjohnson:
To be released with: https://github.com/Hylozoic/hylo-evo/pull/135

### DIFF
--- a/api/models/linkPreview/findOrCreateByUrl.js
+++ b/api/models/linkPreview/findOrCreateByUrl.js
@@ -2,14 +2,6 @@ export default function findOrCreateByUrl (url) {
   return LinkPreview.find(url).then(preview => {
     if (!preview) return LinkPreview.queue(url)
     if (!preview.get('done')) return
-    return preview.pick(
-      'id',
-      'url',
-      'image_url',
-      'title',
-      'description',
-      'image_width',
-      'image_height'
-    )
+    return preview
   })
 }
--- a/api/models/linkPreview/findOrCreateByUrl.js
+++ b/api/models/linkPreview/findOrCreateByUrl.js
@@ -11,8 +11,17 @@ export default function (url) {
       return {status: 'loading'}
     }
 
-    return preview.pick(
-      'id', 'url', 'image_url', 'title', 'description', 'image_width', 'image_height'
+    return Object.assign(
+      preview.pick(
+        'id',
+        'url',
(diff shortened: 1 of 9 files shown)


api/models/linkPreview/findOrCreateByUrl.js
```diff
@@ -0,0 +1,27 @@
...
+      preview.pick(
+        'id',
+        'url',
+        'image_url',
+        'title',
+        'description',
+        'image_width',
+        'image_height'
+      ),
+      {status: 'loaded'}
```
levity: Since this value is never used by the front-end, all you really need to do here is return the `preview` instance.

lorenjohnson: Ah yes, this is vestigial to earlier work. Will move it out.

lib/graphql-bookshelf-bridge/makeResolvers.js
```diff
@@ -15,7 +15,7 @@ export function createResolverForModel (spec, fetcher) {
     transform(attributes, (result, attr) => {
       result[camelCase(attr)] = x => {
         if (typeof x[attr] === 'function') return x[attr]()
-        return x[attr] || x.get(attr)
+        return x[attr] || (x.get && x.get(attr))
```
levity: realized there's a bug here predating your change -- will add a fix to this branch

lib/graphql-bookshelf-bridge/makeResolvers.js
```diff
@@ -86,3 +81,11 @@ export function createResolverForModel (spec, fetcher) {
     }, {})
   )
 }
+
+export function resolveAttribute (result, attr) {
```
lorenjohnson: 👍

api/models/linkPreview/findOrCreateByUrl.js
```diff
@@ -1,27 +1,15 @@
...
-        'image_url',
-        'title',
-        'description',
-        'image_width',
-        'image_height'
-      ),
-      {status: 'loaded'}
+    if (!preview) return LinkPreview.queue(url)
+    if (!preview.get('done')) return
+    return preview.pick(
```
levity: We shouldn't call `pick` here, because this is at the model layer. GraphQL will take care of filtering down what's returned to the client.

PR #304, opened 2017-07-05
Update post count

robbiecarlton:


api/models/Membership.js
```diff
@@ -91,7 +91,10 @@ module.exports = bookshelf.Model.extend(merge({
     return bookshelf.knex('communities_users').where({
       user_id: user_id,
       community_id: community_id
-    }).update({last_viewed_at: new Date()})
+    }).update({
+      last_viewed_at: new Date(),
+      new_post_count: 0
```
robbiecarlton: This is happening in a query, not a mutation, but I think the same argument for updating last_viewed_at applies here.

levity:
Why not reuse what was set up for when the home link is clicked?

robbiecarlton:
hmm. It makes sense to me to clear newPostCount any time you're fetching posts for the community. Otherwise if you navigate straight to /c/foo, rather than click a link in the app, it won't clear the count.
If we really want to keep those kind of mutations out of queries, we could do it by calling resetPostCount in the Feed lifecycle methods

PR #306, opened 2017-07-08
Network graphql

lorenjohnson:
Tests remain


api/graphql/makeModels.js
```diff
@@ -198,10 +201,7 @@ export default function makeModels (userId, isAdmin) {
             sort: sortBy,
             topic
           })
-      },
-      filter: nonAdminFilter(relation => relation.query(q => {
```
levity: Removing this filter would give a user access to all the communities in the system through the top-level `community()` query. I think what you have to do here instead is keep the filter and do this:

```js
q => {
  q.where(function () {
    this.where('communities.id', 'in', myCommunityIds(userId))
    .orWhere('communities.id', 'in', myNetworkCommunityIds(userId))
  })
}
```

api/models/network/updateNetwork.js
```diff
@@ -0,0 +1,62 @@
+import { isEqual, difference, values, some } from 'lodash'
+import setupNetworkAttrs from './setupNetworkAttrs'
+
+export default function updateNetwork (userId, id, params) {
+  if (!userId) throw new Error('updateNetwork called with no userID')
+  if (!id) throw new Error('updateNetwork called with no ID')
+  return setupNetworkAttrs(userId, params).then(attrs =>
+    bookshelf.transaction(transacting =>
+      // NOTE: EnsureLoad not built to work with belongsToMany relations
```
levity: It should work. What happened when you tried?

In any case you don't really need it here since this code is only good for use on a single network at once. If it were being passed in a Network instance then it would be a different story.

api/models/network/updateNetwork.js
```diff
@@ -0,0 +1,62 @@
...
+    updateModerators(network, moderator_ids && values(moderator_ids), transacting) // eslint-disable-line camelcase
+  ])
+}
+
+export function updateCommunities (network, newCommunityIds, transacting) {
+  if (!newCommunityIds) return
+  const currentCommunityIds = network.relations.communities.pluck('id')
+  if (!isEqual(newCommunityIds, currentCommunityIds)) {
+    const communitiesToAdd = difference(newCommunityIds, currentCommunityIds)
+    const communitesToRemove = difference(currentCommunityIds, newCommunityIds)
```
levity: typo here

api/models/network/updateNetwork.js
```diff
@@ -0,0 +1,62 @@
...
+export function afterUpdatingNetwork (network, opts) {
+  const {
+    params: {
+      community_ids,
+      moderator_ids
+    },
+    transacting
+  } = opts
+  return Promise.all([
+    updateCommunities(network, community_ids && values(community_ids), transacting), // eslint-disable-line camelcase
```
levity: i think `values` here is a no-op because these are already arrays... ?

api/models/network/updateNetwork.js
lorenjohnson: Hmm. It seems cleared inspecting the relevant code at the time. Mixing-in EnsureLoad and using network.ensureLoad(['communities', 'moderators']) as done in updatePost was all I was after, but it wouldn't have made much of a difference all told. I haven't grok'd yet why this isn't relevant here in the same way it was in updatePost? Doesn't matter much for now though.

api/models/network/updateNetwork.js
lorenjohnson: Yeah, well actually this was a bug that I hadn't tracked-down and I'm glad you found it again. `community_ids` is actually coming in as an index object (`{1: <someCommunityId>,  2: <anotherCommunityId>, ... }`) at this stage for some reason. I need to go back and track down why.

api/graphql/makeModels.js
lorenjohnson: Will fix. @robbiecarlton started off on the ticket and it was his fault. It wasn't me :)

api/models/network/updateNetwork.js
levity: it's present in `afterUpdatingPost` because that function is being passed an instance, not an ID. So it could be useful in `afterUpdatingNetwork` as well, but not in `updateNetwork`. Perhaps that's what you were intending, but I was responding to the placement of the comment.

api/graphql/makeModels.js
lorenjohnson: Solved this as you recommend but needed to wrap these in an add'l `.where(q => ...` (AND) to get the right results. Works fine.

lorenjohnson:
@levity this should be ready for re-review. Corrected all the things.

levity:
Can you take the comment out about EnsureLoad? Other than that, :+1:

PR #308, opened 2017-07-17
Network communities

robbiecarlton:


api/controllers/CommunityController.js
```diff
@@ -228,14 +228,18 @@ module.exports = {
...
       user_id: req.param('userId'),
       community_id: req.param('communityId')
     }).update({
       active: false,
       deactivated_at: new Date(),
       deactivator_id: req.session.userId
     })
+    .then(() => Community.find(req.param('communityId'))
+      .then(community => community.save({
+        num_members: community.get('num_members')
```
levity: i think you want a `- 1` here

but then, to skip the 2 queries of "find then save", you should use [decrement](http://knexjs.org/#Builder-decrement)

api/models/Membership.js
```diff
@@ -58,6 +58,9 @@ module.exports = bookshelf.Model.extend(merge({
     .save({}, {transacting: opts.transacting})
     .tap(() => User.followDefaultTags(userId, communityId, opts.transacting))
     .tap(() => Community.find(communityId, {transacting: opts.transacting})
+      .tap(community => community.save({
+        num_members: community.get('num_members') + 1
```
levity: i guess not using [increment](http://knexjs.org/#Builder-increment) here is ok because you already have the community instance... but one thing to be aware of is that by default, the save operation will create a query that updates all the fields in the model, not just the one you changed. if you just want that, you have to use `patch`:

```js
community.save({num_members: 7}, {patch: true, transacting}) // note it's in a separate object
```

lorenjohnson:
FYI -- I kind need/want this (memberCount in particular) for my Network Members search work now. I already started making some of these changes locally but hadn't pushed anything... Maybe I'll merge it into my branch and keep it up to date with your changes and you finalize.

robbiecarlton:
@levity - made those changes, good to merge?

PR #310, opened 2017-07-19
Web 35 worker error stacktraces

razorman8669:
@richchurcher I went through and tested this thoroughly and stacktraces appear to be spitting out correctly in all cases now.  I also added the jobId to the rollbar handling.  was there anything else you were thinking about in for this issue: https://hylozoic.atlassian.net/browse/WEB-35?

### DIFF
--- a/cron.js
+++ b/cron.js
@@ -85,6 +85,7 @@ skiff.lift({
     })
     .catch(function (err) {
       sails.log.error(red(err.message))
+      sails.log.error(err)
       rollbar.handleError(err, () => skiff.lower())
     })
   }
--- a/worker.js
+++ b/worker.js
@@ -52,7 +52,7 @@ var processJobs = function () {
         const error = typeof err === 'string'
           ? new Error(err)
           : (err || new Error('kue job failed without error'))
-        sails.log.error(label + error.message.red)
+        sails.log.error(label + error.message.red, error)
         rollbar.handleErrorWithPayloadData(error, data)
         done(error)
       })
--- a/worker.js
+++ b/worker.js
@@ -48,7 +48,7 @@ var processJobs = function () {
         done()
       })
       .catch(err => {
-        const data = {custom: {jobData: job.data}}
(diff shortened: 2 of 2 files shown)


razorman8669:
@richchurcher i'm going to merge this in once you response.  thanks

richchurcher:
@razorman8669 So the issue is that ordinary exception stack traces are not displayed on jobs (which run using Kue). To reproduce this, take for example Notifications which is where this all started becoming painful:
```
      case MEDIUM.InApp:
        // console.log(iDoNotExist)
        // throw new Error('HI I AM AN ERROR, DO NOT SWALLOW ME!')
        const userId = this.reader().id
        action = User.incNewNotificationCount(userId)
          .then(() => this.updateUserSocketRoom(userId))
        break
```
If I uncomment either of the commented out lines, nothing will show up in the worker's output. As you can imagine this makes debugging fairly painful, since even a typo can go unnoticed for ages.

I suspect this is because we're not using Kue's y`error` event](https://github.com/Automattic/kue#error-handling) although haven't tested that theory.

razorman8669:
@richchurcher I tried reproducing your example and the code I committed does indeed spit out the error to the console as well as logs to rollbar.  Maybe we should pair up to see if there is something different going on.

richchurcher:
That is definitely strange. I wonder what the difference is in our environments? Are you triggering that code using an @mention? Yeah let's jump into a Zoom together at some point and figure it out :smile:

levity:
I've been noticing an uptick in "Unhandled error" notices from Logentries. I hope this work resolves that as well.

richchurcher:
Yeah, I'm suspicious it's my `console.error` in the notifications code, since I feel like it started around the same time. I'll see if I can ponder out something...

richchurcher:
I see no harm in merging this, although I still can't get local logging of some errors to work. I also suspect we need to handle redis error events but I'll deal with that in a separate PR. Thanks @razorman8669  and sorry to keep you waiting so long!

PR #311, opened 2017-07-20
Skills backend

robbiecarlton:


api/models/Skill.js
```diff
@@ -0,0 +1,19 @@
...
+    return this.belongsToMany(User, 'skills_users')
+  }
+
+}, {
+
+  find: function (nameOrId, opts = {}) {
+    if (!nameOrId) return Promise.resolve(null)
+
+    const where = isNaN(Number(nameOrId))
+      ? {name: nameOrId}
```
levity: this should be case-insensitive

PR #315, opened 2017-07-27
Skills autocomplete web 87

robbiecarlton:


api/graphql/filters.js
```diff
@@ -69,3 +69,13 @@ export function communityTopicFilter (userId, {
     }
   }
 }
+
+export const skillInCommunitiesFilter = curry((userId, relation) =>
```
lorenjohnson: I know it's verbose but I'd call this skillInCommunitiesOrNetworksFilter

PR #317, opened 2017-07-28
Search users by skills web 48

robbiecarlton:


robbiecarlton:
@lorenjohnson so this does it, with a little ugliness.
Specifically, the fact that we're using two different methods to fetch users, depending on whether there is a search or not.

However, it does do everything we need it to do, efficiently. My inclination is to merge this, and push forward with the skills work, and then get @levity's take when he's back. Feels like there's a significant refactor brewing here that I would want his input on.

levity:
What I would have suggested here is to move this "use FullTextSearch when there's a search term; use the other implementation otherwise" logic into `services/Search/forUsers`. `searchQuerySet` is just a wrapper around that, after all.

Yes, this technically changes the existing behavior of code that uses Search.forUsers with a `term` option set, but that's fine.

PR #325, opened 2017-08-23
Members Search works with prefixes WEB-105

robbiecarlton:


robbiecarlton:
@levity - made a couple more changes to fix tests, would you take one more look?

PR #328, opened 2017-08-31
DEV-57: file upload handling (WIP)

levity:


api/services/AssetManagement.js
```diff
@@ -1,68 +1,51 @@
...
-const request = require('request')
-
+import crypto from 'crypto'
+import path from 'path'
+import Promise from 'bluebird'
+import request from 'request'
+import sharp from 'sharp'
+import { createS3StorageStream } from './Uploader/storage'
+
+// this defends against filenames that are blank or have troublesome characters
```
razorman8669: @levity i've used this https://www.npmjs.com/package/sanitize-filename in the past.  could be useful...

api/services/Uploader.js
```diff
@@ -0,0 +1,107 @@
...
+    const type = fileType(firstDataChunk)
+    if (type) {
+      finalFilename = finalFilename.replace(/(\.\w{2,4})?$/, '.' + type.ext)
+    }
+  } catch (err) {}
+
+  return finalFilename
+}
+
+if (require.main === module) {
```
razorman8669: what's the purpose of this?

api/services/Uploader.js
```diff
@@ -0,0 +1,107 @@
...
+  const dotenv = require('dotenv')
+  dotenv.load()
+
+  const arg = process.argv[2]
+
+  upload({
```
razorman8669: and why are you hardcoding userId 42 and type `userBanner`?

api/services/Uploader.js
levity: oh yeah, that line means "run this if the current file is being run as a script". i was using that for testing. i'll take it out

PR #333, opened 2017-09-11
DEV-81: File uploads pt. 2

levity:


levity:
hmm -- well, it doesn't depend on any models except for in validation. which suggests a later refactoring to allow the validation to be passed into the uploader from the models side...

PR #334, opened 2017-09-11
Search for community by slug

andygimma:


api/graphql/index.js
```diff
@@ -59,6 +59,12 @@ function createSchema (userId, isAdmin) {
             }
           })
       },
+      communityExists: (root, { slug }) => {
+        return Search.forCommunities({ slug }).fetch().then(result => {
```
levity: Never call `fetch` if all you care about is existence or count, because Bookshelf will do unnecessary work to prepare the instance(s). This suffices:

```js
Community.where(bookshelf.knex.raw('lower(slug) = lower(?)', slug)).count()
```

Note that you have to first convert the argument supplied by the user to a valid slug, i.e. remove invalid characters and convert to lowercase.

api/graphql/schema.graphql
```diff
@@ -316,7 +316,12 @@ type VoteQuerySet {
   items: [Vote]
 }
 
+type CommunityExistsOutput {
+  data: Boolean
```
levity: please name this something meaningful like `exists`

api/graphql/index.js
andygimma: Are any characters other than numbers, letters, and dashes valid? I didn't see an existing validator so I think I'll add one to the Community model.

richchurcher: @andygimma If it helps, I just added a rather raw validators infrastructure to hylo-utils. You could add a `validateCommunity` object to that, with a `slug` function.

Here's `generateSlug` from hylo-redux, which could be converted to String.match to validate:
```
const generateSlug = (name, oldSlug) => {
  if (oldSlug) {
    return incrementSuffix(oldSlug)
  } else {
    if (!name) return ''
    return name.toLowerCase().substring(0, 40).trim()
    .replace(/\s/g, '-')
    .replace(/[\!\?'"\/\\\.,]/g, '')
  }
}
```

PR #337, opened 2017-09-12
DEV-65 - Add simple password and name validation

richchurcher:
Depends on https://github.com/Hylozoic/hylo-utils/pull/4 . Will update package.json with the correct hash after that gets merged.

This does simple checks to avoid empty or short passwords, and empty names. 'password' gets temporarily added to the whitelist to keep all the validation in the same function.


richchurcher:
The CircleCI build will fail until hylo-utils is updated :)

richchurcher:
Actually, some older UserController tests are failing locally so I need to track that down first!

PR #338, opened 2017-09-12
Join Community by Invite - WEB-81

lorenjohnson:
https://hylozoic.atlassian.net/browse/WEB-81

A front-end PR is still in the works, but this one should be good to merge.


api/models/Invitation.js
```diff
@@ -47,7 +47,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
     const { transacting } = opts
     return Membership.where({
       user_id: userId, community_id: this.get('community_id')
-    }).fetch()
+    }).fetch({withRelated: 'community'})
```
levity: can you add a comment here about how this is so that the data is available for callers, even though it's not used in the code below?

lorenjohnson:
@levity I will do both and also note that I've just pushed an additional graphql query endpoint for checking invite validity. You might want to take a quick look at that addition.

api/graphql/index.js
```diff
@@ -89,7 +89,14 @@ function createSchema (userId, isAdmin) {
         }),
       network: (root, { id, slug }) =>  // you can specify id or slug, but not both
         fetchOne('Network', slug || id, slug ? 'slug' : 'id'),
-      skills: (root, args) => fetchMany('Skill', args)
+      skills: (root, args) => fetchMany('Skill', args),
+      checkInvitation: (root, { invitationToken }) =>
+        Invitation.query()
+        .where(bookshelf.knex.raw('token = ? AND used_by_id IS NULL', invitationToken))
```
levity: i don't think you need `raw` for this; this should work:
```js
.where({token: invitationToken, used_by_id: null})
```

lorenjohnson:
All changes applied. The raw came from a misunderstanding of what Andy had recommended given some feedback you'd given him about making an existence query...

levity:
looks like there are still conflict markers in yarn.lock -- maybe just merge master and rerun `yarn install` and check in whatever you get?

lorenjohnson:
@levity good for final review

levity:
i overlooked that there were still merge conflicts in yarn.lock :( fixed in master

lorenjohnson:
Ugh. Yeah, I saw that after merging. It’s weird. I corrected it 3 times including re-yarning on my branch and committing my lock… Bleh. Thanks

Loren

PR #340, opened 2017-09-14
DEV-75 mutation validations on skill and some code coverage for mutat…

razorman8669:
…ions


api/graphql/mutations/index.js
```diff
@@ -144,6 +145,11 @@ export function deletePost (userId, postId) {
...
 
 export function addSkill (userId, name) {
+  if (isEmpty(trim(name))) {
+    throw new Error('Skill cannot be blank')
+  } else if (size(name) > 39) {
+    throw new Error('Skill must be less than 40 characters')
+  }
   return Skill.find(name)
   .then(skill => {
     if (!skill) return new Skill({name}).save()
```
levity: I realize that this was here before you started working on it, but we should also find and save the skill using the trimmed version. Can you take care of that and then merge?

levity:
oh, you're off today, right? i'll finish this up then.

PR #344, opened 2017-09-18
DEV-80: validate flagInappropriateContent fields

richchurcher:
https://hylozoic.atlassian.net/browse/DEV-80. Depends on https://github.com/Hylozoic/hylo-utils/pull/5.

  - [x] Checks `category` to match a subset I pretty much invented. If we have this list somewhere let me know... otherwise I guess we can refine them later.
  - [x] sense-checks `link` to make sure it's from us
  - [x] prevents extremely large chunks of text in `reason`


richchurcher:
As before, will add hylo-utils hash to package.json after merge :smile:

api/models/FlaggedItem.js
```diff
@@ -1,8 +1,35 @@
...
+      return Promise.reject('Unknown category.')
+    }
+
+    const invalidReason = validateFlaggedItem.reason(reason)
+    if (invalidReason) Promise.reject(invalidReason)
+
+    const invalidLink = validateFlaggedItem.link(link)
+    if (invalidLink) return Promise.reject(invalidLink)
+
+    return FlaggedItem.forge(attrs).save()
```
levity: it's safe to use `this` here instead of `FlaggedItem`, i believe -- i know it doesn't matter much, but it feels better to use a local reference than a global one where possible

richchurcher: Will do!

richchurcher:
NB: for all the unused deps removed, I did the following:
```
ag "require\('depname'\)"
ag "from 'depname'"
```
then checked git blame to see when they were included, and what the parent commits were. Pretty sure these are safe to go.

levity:
probably best to add the hylo-utils change to package.json in this PR so that we can see circleci get through it successfully

richchurcher:
Oh definitely, I wasn't going to try to commit until after everything went green.

PR #348, opened 2017-09-21
disabling push notifications (DEV-100)

levity:


api/models/PushNotification.js
```diff
@@ -14,9 +14,11 @@ module.exports = bookshelf.Model.extend({
     var alert = this.get('alert')
     var path = this.get('path')
     var badgeNo = this.get('badge_no')
+    const disabled = !!process.env.DISABLE_PUSH_NOTIFICATIONS
 
-    return this.save({'sent_at': (new Date()).toISOString()}, options)
-    .then(pn => OneSignal.notify(platform, deviceToken, alert, path, badgeNo))
+    return this.save({sent_at: (new Date()).toISOString(), disabled}, options)
```
robbiecarlton: so you're still creating the Push Notification, but then saving it with a disabled flag. Just so we have a record of everything?

levity: yep!

PR #350, opened 2017-09-21
Support for Join Community BetaAccessCode (WEB-217)

lorenjohnson:


api/services/InvitationService.js
```diff
@@ -137,12 +137,57 @@ module.exports = {
...
-  use: (userId, invitationToken) => {
-    return Invitation.where({token: invitationToken}).fetch()
-    .then(invitation => {
-      if (!invitation) throw new Error('not found')
-      if (invitation.isExpired()) throw new Error('expired')
-      return invitation.use(userId)
-    })
+  check: (userId, token, betaAccessCode) => {
+    if (betaAccessCode) {
+      return Community.query(qb => {
```
levity: please de-dupe the code that's here and also on line 164

lorenjohnson: will do

lorenjohnson:
@levity this is ready for your eyes again. I have added tests to the underlying InvitationService and Community model, I've not added anything for the graphql endpoint and may need to, but want your guidance / metric on that one this round.

levity:
I'm definitely still figuring out how to test the graphql endpoints well. The current tests in api/graphql/index.test.js definitely feel a bit too heavy. It's good to have a few ones that go end-to-end but I think the typical test should be more unit-focused. Definitely a work in progress and something I'll be looking for feedback from the whole team on.

richchurcher:
@levity a lot of our backend tests are what I think of as integration tests, involving many different units of code (and usually a database hit). It would be really interesting to try to understand how we could corral those into a more defined subset, and do mostly 'pure' unit testing with fewer integration tests and some E2E. I say 'interesting' in the knowledge that it's actually quite a lot of work :sunglasses:

PR #352, opened 2017-09-22
MBL-77 Allow div in sanitize

richchurcher:
As it turns out, the comment editor newline bug was twofold: `sanitize` was stripping it out in the component and in the backend. Updated `hylo-utils` again (gulp) so this is the backend part. It does definitely need complete removal of `node_modules`, `yarn cache clean` etc. It doesn't seem to need `yarn upgrade hylo-utils` though.


richchurcher:
Closing in favour of https://github.com/Hylozoic/HyloReactNative/pull/59.

PR #354, opened 2017-09-22
New push notification setup part 1 (DEV-55)

levity:


api/models/PushNotification.js
```diff
@@ -73,5 +76,9 @@ module.exports = bookshelf.Model.extend({
 
   textForApprovedJoinRequest: function (community, actor) {
     return `${actor.get('name')} approved your request to join ${community.get('name')}`
+  },
+
+  setupDeviceIds: function () {
```
robbiecarlton: empty function, is this supposed to be here?

levity: ah good catch

PR #355, opened 2017-09-25
New push notification setup part 2 (DEV-55)

levity:


levity:
you don't even have to pass it anything; you just call a method in the OneSignal lib and it gives you a player ID. Presumably it gets the token and passes it up behind the scenes.

PR #356, opened 2017-09-29
MBL-17 update logic for flagging content mutations to generate links on backend

razorman8669:
This changes the logic of the flagInappropriateContent logic so that links are generated on the backend instead of passing the link from the frontend (ie: mobile or web).

The PR for the frontend is here: https://github.com/Hylozoic/HyloReactNative/pull/67
(though it's not finished yet).

### DIFF
--- a/api/graphql/mutations/index.js
+++ b/api/graphql/mutations/index.js
@@ -132,7 +132,22 @@ export function removeSkill (userId, skillIdOrName) {
   .then(() => ({success: true}))
 }
 
-export function flagInappropriateContent (userId, { category, reason, link }) {
+export function flagInappropriateContent (userId, { category, reason, linkData }) {
+  let link
+  switch (trim(linkData.type)) {
+    case 'post':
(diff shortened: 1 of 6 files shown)


api/graphql/mutations/index.js
```diff
@@ -132,7 +132,22 @@ export function removeSkill (userId, skillIdOrName) {
   .then(() => ({success: true}))
 }
 
-export function flagInappropriateContent (userId, { category, reason, link }) {
+export function flagInappropriateContent (userId, { category, reason, linkData }) {
+  let link
```
levity: add some tests for this please -- if you want to pull it out into its own function and just test that instead of going through FlaggedItem.create each time, that'd be good too. (if so, might want to move the two functions into `api/graphql/mutations/flagging.js`

razorman8669: @levity done

razorman8669:
@levity requested tests implemented.  please approve if good

PR #357, opened 2017-10-09
bump node version to 8.6

levity:
also use nodemon instead of forever


Procfile.dev
```diff
@@ -1,2 +1,2 @@
-web: forever -c node --minUptime 1000 --spinSleepTime 10000 -w app.js
-worker: DELAY_START=5 forever -c node --minUptime 1000 --spinSleepTime 10000 -w worker.js
+web: nodemon app.js
```
razorman8669: can you change this to `nodemon --debug=3001 app.js --delay 0.5`
The debug is so we can attach a remote debug sessions and step through code (AFAIK it doesn't add performance drain when a debugger isn't attached).
the `delay` is because I've seen with nodemon that it will trigger multiple rebuilds sometimes when ide's do `SAVE ALL` actions and it affects multiple files.

Procfile.dev
```diff
@@ -1,2 +1,2 @@
...
+worker: DELAY_START=5 nodemon worker.js
```
razorman8669: add delay here too

Procfile.dev
levity: ok. does the order of arguments matter here?

razorman8669: no

PR #359, opened 2017-10-11
MBL-81: set backend up to deliver socket updates in correct format 

richchurcher:
This:
  - [x] returns a slightly modified shape from `updateUserSocketRoom` which removes the guesswork I used when I wrote it the first time around
  - [x] returns a much more detailed post object from `createPost`
  - [x] migrates `link_preview_id` from `integer` to `bigint`
  - [x] adds some helpers to `api/models/util` to simplify some of the boilerplate

### DIFF
--- a/api/models/Notification.js
+++ b/api/models/Notification.js
@@ -298,19 +298,30 @@ module.exports = bookshelf.Model.extend({
   },
 
   updateUserSocketRoom: function (userId) {
+    const { activity } = this.relations
     const actor = Object.assign({},
       this.actor().pick([ 'id', 'name' ]),
       { avatarUrl: this.actor().get('avatar_url') }
     )
+    const meta = activity.get('meta')
+    const action = Notification.priorityReason(meta.reasons)
     const comment = this.comment().pick([ 'id', 'text' ])
-    const community = this.relations.activity.relations.community.pick([ 'id', 'name', 'slug' ])
+    const community = activity.relations.community.pick([ 'id', 'name', 'slug' ])
+    const createdAt = activity.get('created_at').toString()
     const post = { id: this.post().id, title: this.post().get('name') }
     const payload = {
       id: '' + this.id,
-      createdAt: this.get('created_at'),
-      activity: Object.assign({},
-        this.relations.activity.pick([ 'action', 'id', 'meta', 'unread' ]),
-        { actor, comment, community, post })
+      activity: {
+        action,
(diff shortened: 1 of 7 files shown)


levity:
won't this affect code in evo that's expecting the current shape?

levity:
never mind -- i just tried it in evo and the changes are minor enough that they don't disrupt any existing code

richchurcher:
@levity Yeah the changes are modelled off the graphql request we make in evo, which I used essentially unchanged for mobile... so I think we're ok. Plus I don't think socket updates are wired in yet for evo notifications.

I do have one possible addition/change for the `newNotificationsCount` so I'll hold off merging for a bit.

api/models/Notification.js
```diff
@@ -298,19 +298,30 @@ module.exports = bookshelf.Model.extend({
...
       this.actor().pick([ 'id', 'name' ]),
       { avatarUrl: this.actor().get('avatar_url') }
     )
+    const meta = activity.get('meta')
+    const action = Notification.priorityReason(meta.reasons)
     const comment = this.comment().pick([ 'id', 'text' ])
-    const community = this.relations.activity.relations.community.pick([ 'id', 'name', 'slug' ])
+    const community = activity.relations.community.pick([ 'id', 'name', 'slug' ])
+    const createdAt = activity.get('created_at').toString()
     const post = { id: this.post().id, title: this.post().get('name') }
```
levity: minor quibble -- if this line is only used in the final payload object, i'd rather just see it listed in-line there. but up to you

richchurcher:
@levity time for another look? Obviously needs tests.

api/models/Post.js
```diff
@@ -259,8 +260,40 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+        commentsTotal: 0,
+        communities: refineMany(communities, [ 'id', 'name', 'slug' ]),
+        creator,
+        linkPreview: refineOne(linkPreview, [ 'id', 'image_url', 'title', 'url' ]),
+        topics,
+
+        // May need to retain these to avoid breaking legacy code.
+        // TODO: Check if these are still required.
+        creatorId: creator.id,
+        tags: topics
```
levity: Yep, these two are still required in SocketListener on evo: https://github.com/Hylozoic/hylo-evo/blob/master/src/components/SocketListener/SocketListener.store.js#L55

PR #360, opened 2017-10-11
Mbl 98 force update

andygimma:


api/controllers/MobileAppController.js
```diff
@@ -1,8 +1,22 @@
 module.exports = {
-
   updateInfo: function (req, res) {
-    var result = {}
```
levity: Since the new implementation is entirely in the other method, you can remove this method and the routes.js entry associated with it completely. The old apps are no longer accessing it because they just get redirected to hylo.com/newapp.

api/controllers/MobileAppController.js
```diff
@@ -12,26 +26,56 @@ module.exports = {
...
 
-    res.ok(result)
+    return res.ok(result)
+  }
+}
+
+function resultBuilder (type, platform) {
+  var iTunesItemIdentifier = '1002185140'
+  var appStoreLink = `https://itunes.apple.com/app/id${iTunesItemIdentifier}`
+  var playStoreLink = 'https://play.google.com/store/apps/details?id=com.hylo.reactnative'
```
levity: please have these two urls be environment variables: IOS_APP_STORE_URL and ANDROID_APP_STORE_URL

api/controllers/MobileAppController.js
```diff
@@ -12,26 +26,56 @@ module.exports = {
...
-      case '1.65':
-        result = {
-          type: 'force',  // can be 'force' or 'suggest'
-          title: 'A new version of the app is available',
-          message: 'The version you are using is no longer compatible with the site. Please go to the App Store now to update',
-          iTunesItemIdentifier: '1002185140'
-        }
+    switch (req.param('ios-version')) {
+      case '2.0':
+        result = resultBuilder(SUGGEST, IOS)
```
levity: you want to let the 2.0 versions pass through without interruption, no? and basically force everything else?

PR #361, opened 2017-10-11
Update profile link with Facebook token auth

robbiecarlton:


api/models/LinkedAccount.js
```diff
@@ -60,7 +60,7 @@ module.exports = bookshelf.Model.extend({
     switch (type) {
       case 'facebook':
         return {
-          facebook_url: profile.profileUrl,
+          facebook_url: profile.profileUrl || profile._json.link,
```
levity: might want to use `get` here in case neither are defined

levity:
huh, and your diff-coverage works.

PR #363, opened 2017-10-18
remove old code

levity:
TODO

- [x] remove PostPresenter and CommentPresenter
  - requires modifying SocketListener on evo and RN to handle new data structure, because the last place using the presenters is createComment


config/routes.js
```diff
@@ -1,117 +1,21 @@
-/* eslint key-spacing:0 */
+/* eslint-disable key-spacing */
```
lorenjohnson: Yay! This gutting of the routes file really highlights the gain in the reduced code and noise/complexity in the app.

PR #367, opened 2017-11-01
quotes looked like &quot; in slack messages

Connoropolous:
They didn't used to so I'm not sure when it changed.

screenshot_20171101-104027


levity:
Thanks @Connoropolous!

levity:
FWIW though this is not happening with me on the macOS Slack app, so I think the culprit here is your Android Slack app.

Connoropolous:
ah, interesting

PR #369, opened 2017-11-16
Force update

andygimma:


api/controllers/MobileAppController.js
```diff
@@ -1,6 +1,11 @@
 import semver from 'semver'
 
 module.exports = {
+  updateInfo: function (req, res) {
+    const iosVersion = req.param('ios-version')
+    const platform = iosVersion ? IOS : ANDROID
+    return res.ok(resultBuilder(FORCE, platform))
```
robbiecarlton: I don't think `resultBuilder` works here. See https://github.com/Hylozoic/hylo-node/pull/360/files
The old ios client is expecting the endpoint to return `iTunesItemIdentifier ` as part of the payload.
I just looked over the old Android code and afaict that app never calls this endpoint, so it doesn't matter what you return for Android.

PR #371, opened 2017-11-16
Order pinned posts (part 2) DEV-133

robbiecarlton:


robbiecarlton:
@levity - good to merge now?

api/graphql/makeModels.js
```diff
@@ -432,14 +432,12 @@ export default function makeModels (userId, isAdmin) {
     PostMembership: {
       model: PostMembership,
       attributes: [
-        'id'
+        'id',
+        'pinned'
```
levity: i recently discovered that this is not necessary at all: https://www.apollographql.com/docs/graphql-tools/resolvers.html#Default-resolver

so just remove that and you're good to merge

PR #372, opened 2017-11-18
DEV-140 -- Eliminate generation of duplicate entries to networks_posts on post creation

lorenjohnson:
** Includes migration to remove duplicates from ``networks_posts``


migrations/20171117161044_delete_networks_posts_dupes.js
```diff
@@ -0,0 +1,15 @@
+exports.up = knex => knex.raw(`
+  DELETE FROM networks_posts
+  WHERE id IN (
+    SELECT id
+    FROM (
+      SELECT
+        id,
+        ROW_NUMBER() OVER(PARTITION BY network_id, post_id ORDER BY id ) AS row_num
```
levity: 🆒

lorenjohnson: Indeed. This was like the bug fix that epitomizes what I THINK most bug fixes are going to be but often aren't 👍

PR #375, opened 2017-11-29
Mbl 81/socket redis testing

richchurcher:


richchurcher:
Investigating the redis thing on Circle.

richchurcher:
Ok @levity I think we're all working now...

circle.yml
```diff
@@ -1,6 +1,10 @@
 machine:
+  environment:
+    REDIS_URL: redis://0.0.0.0:6379/1
   node:
     version: 8.6.0
+  services:
+    - redis
```
levity: well isn't that pleasingly simple.

richchurcher: I know! If we didn't require a host in the Node source, I doubt I would even have had to provide an environment variable.

PR #377, opened 2017-11-30
Web 330/network settings mutations

richchurcher:
This:
 - [x] adds mutations for single add/remove of moderators and communities
 - [x] makes `isAdmin` available in `makeMutations`
 - [x] reformats `makeMutations` so I can read it (lemme know if you hate all that whitespace!)

### DIFF
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -4,6 +4,8 @@ import { join } from 'path'
 import setupBridge from '../../lib/graphql-bookshelf-bridge'
 import { presentQuerySet } from '../../lib/graphql-bookshelf-bridge/util'
 import {
+  addCommunityToNetwork,
+  addNetworkModeratorRole,
   addModerator,
   addSkill,
   createComment,
@@ -23,8 +25,10 @@ import {
   regenerateAccessCode,
   registerDevice,
   reinviteAll,
+  removeCommunityFromNetwork,
   removeMember,
   removeModerator,
+  removeNetworkModeratorRole,
   removePost,
   removeSkill,
   resendInvitation,
@@ -54,7 +58,7 @@ async function createSchema (userId, isAdmin) {
 
   const allResolvers = Object.assign({
     Query: makeQueries(userId, fetchOne, fetchMany),
-    Mutation: makeMutations(userId),
+    Mutation: makeMutations(userId, isAdmin),
 
     FeedItemContent: {
       __resolveType (data, context, info) {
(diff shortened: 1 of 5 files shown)


api/graphql/index.js
```diff
@@ -134,52 +138,108 @@ export function makeQueries (userId, fetchOne, fetchMany) {
...
 }
 
-export function makeMutations (userId) {
+export function makeMutations (userId, isAdmin) {
   return {
-    updateMe: (root, { changes }) => updateMe(userId, changes),
-    createPost: (root, { data }) => createPost(userId, data),
-    updatePost: (root, args) => updatePost(userId, args),
+    addCommunityToNetwork: (root, { communityId, networkId }) =>
+      addCommunityToNetwork({ userId, isAdmin }, { communityId, networkId }),
```
levity: wow, you did what i was imagining with the function signature. nice

api/graphql/mutations/network.js
```diff
@@ -2,12 +2,49 @@ import validateNetworkData from '../../models/network/validateNetworkData'
 import underlyingUpdateNetwork from '../../models/network/updateNetwork'
 import convertGraphqlData from './convertGraphqlData'
 
-export function updateNetwork (userId, { id, data }) {
+// TODO: more integrated `isAdmin` handling for mutations?
+export function networkMutationPermissionCheck ({ userId, isAdmin = false }, networkId) {
```
levity: feel free to use async/await btw; i think that would make this simpler?

richchurcher: I saw you started using that with `makeModels`! I'll give it a crack. I also realised I need to check `hasModeratorRole` for the new mod to avoid dups.

api/graphql/mutations/network.js
```diff
@@ -2,12 +2,50 @@ import validateNetworkData from '../../models/network/validateNetworkData'
...
+    throw new Error("You don't have permission to modify that network.")
+  }
+}
+
+export async function addCommunityToNetwork (user, { communityId, networkId }) {
+  await networkMutationPermissionCheck(user, networkId)
+  await Community
+    .where('id', communityId)
+    .save('network_id', networkId, { method: 'update', patch: true })
+  return Network.find(networkId, { withRelated: [ 'communities' ] })
```
richchurcher: @levity See what you think of the above. I suppose it reads well... it feels a little strange, because at each step I'm relying on the underlying code throwing to stop us reaching statements further on in the function. But it's clean-ish? If permissions fail, we throw... if the db operation fails, we throw... and graphql-tools exposes the error in the response.

api/graphql/mutations/network.js
```diff
@@ -2,12 +2,50 @@ import validateNetworkData from '../../models/network/validateNetworkData'
...
+// TODO: more integrated `isAdmin` handling for mutations?
+export async function networkMutationPermissionCheck ({ userId, isAdmin = false }, networkId) {
+  if (isAdmin) return
+  if (!await NetworkMembership.hasModeratorRole(userId, networkId)) {
+    throw new Error("You don't have permission to modify that network.")
+  }
+}
+
+export async function addCommunityToNetwork (user, { communityId, networkId }) {
+  await networkMutationPermissionCheck(user, networkId)
```
levity: I find it misleading to name this first argument `user`, since it's not a user object.

api/graphql/mutations/network.js
levity: Looks fine to me!

api/graphql/mutations/network.js
richchurcher: Yeah, I struggled with that too. I tried 'credentials' but it seemed weird. Here's a try with `authZ` (vs authN), see if that makes sense to you.

levity:
maybe just `auth`? does the `Z` signify something i'm missing?

richchurcher:
@levity Sorry missed this... yep, `authZ` is authoriZation, `authN` is autheNtication. Maybe those terms have lapsed into disuse :) Since this was about privileges I figured `authZ` made sense, but not if it's not easily recognisable!

richchurcher:
I'm going to go ahead and merge this one but feel free to hotfix the name to whatever makes sense to you!

PR #380, opened 2017-12-01
Axolotl chatbot (DEV-155)

levity:


api/services/Websockets.js
```diff
@@ -9,14 +11,26 @@ const validMessageTypes = [
   'newPost'
 ]
 
+export function broadcast (room, messageType, payload, socketToExclude) {
```
levity: After dealing with a ["Too many Redis connections" issue](https://hylozoic.atlassian.net/wiki/spaces/DEV/pages/265420813/Postmortem+Too+many+Redis+connections) it occurs to me that this would be a good opportunity to set up a connection singleton or connection pool...

PR #384, opened 2017-12-06
Moderator flag notice DEV-12

robbiecarlton:


api/models/comment/createComment.js
```diff
@@ -82,6 +82,8 @@ export function pushMessageToSockets (message, thread) {
   const userIds = followers.pluck('id')
   const excludingSender = userIds.filter(id => id !== message.get('user_id'))
 
+  console.log('pushing message with date', message.get('created_at'))
```
levity: remove console.log

api/services/Websockets.js
```diff
@@ -9,14 +11,26 @@ const validMessageTypes = [
   'newPost'
 ]
 
+export function broadcast (room, messageType, payload, socketToExclude) {
```
levity: i thought i left a comment here about reworking this so that it doesn't create a new connection on every call... guess i never submitted it? anyway it doesn't have to block the merge, just noting it.

api/services/MessagingService.js
```diff
@@ -0,0 +1,13 @@
+import findOrCreateThread from '../models/post/findOrCreateThread'
+import createComment from '../models/comment/createComment'
+
+const MessagingService = {
+  async sendMessageFromAxolotl (userIds, text) {
+    // find existing message thread or create a new one
```
levity: this comment is kind of useless now -- i think i just put it there while we were writing this as a temporary reminder

api/services/Websockets.js
levity: although if you wanted to take a stab at this it would be welcome. should be pretty simple, just like
```js
var io

export function broadcast () {
  if (io) {
    io = emitter(process.env.REDIS_URL)
    io.redis.on('error', err => {
      rollbar.error(err, null, {room, messageType, payload})
    })
  }
}
```

robbiecarlton:
@levity - made those changes. Good to merge?

PR #385, opened 2017-12-08
Illegal content notice dev 13

robbiecarlton:
Also requires a change to production environment variables, adding HYLO_ADMINS

### DIFF
--- a/api/models/flaggedItem/notifyUtils.js
+++ b/api/models/flaggedItem/notifyUtils.js
@@ -23,16 +23,15 @@ export async function notifyModeratorsMember (flaggedItem) {
 }
 
 export function sendToCommunities (flaggedItem, communities) {
-  var sendToAdmins = () => Promise.resolve()
-
-  if (flaggedItem.get('category') === FlaggedItem.Category.ILLEGAL &&
-    process.env.HYLO_ADMINS) {
+  const shouldSendToAdmins = flaggedItem.get('category') === FlaggedItem.Category.ILLEGAL &&
+    process.env.HYLO_ADMINS
+  var sendToAdmins = () => {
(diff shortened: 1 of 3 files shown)


api/models/flaggedItem/notifyUtils.js
```diff
@@ -23,11 +23,22 @@ export async function notifyModeratorsMember (flaggedItem) {
 }
 
 export function sendToCommunities (flaggedItem, communities) {
+  var sendToAdmins = () => Promise.resolve()
```
levity: If you're just going to call this within a `.then`, it doesn't need to return a promise because it'll already be wrapped in one. Given that, I would probably rewrite this as:
```js
const shouldSend = ...
// ...
.then(() => shouldSend && sendToAdmins())
```
rather than redefining a function. it seems more straightforward to me that way.

test/unit/models/flaggedItem/notifyUtils.test.js
```diff
@@ -67,7 +67,7 @@ describe('sendToCommunities', () => {
     var expectedUserIds = [modIds1, modIds2]
     var expectedText = [`${message} 1`, `${message} 2`]
 
-    const hyloAdminIds = process.env.HYLO_ADMINS && process.env.HYLO_ADMINS.split(',').map(id => Number(id))
+    const hyloAdminIds = process.env.HYLO_ADMINS.split(',').map(id => Number(id))
```
levity: this will fail on any environment that doesn't have `HYLO_ADMINS` set including a local dev env. better to explicitly define the value to what you want for the purposes of the test rather than imposing required configuration on all test environments.

levity: Then you can also get rid of the `if` wrapper below, so the test then becomes fully deterministic.

PR #386, opened 2017-12-11
Check for empty facebook url object

andygimma:


levity:
before adding a patch here, i would first verify that our own frontend isn't sending an empty object when it should be sending null.

andygimma:
@levity I can confirm that the frontend is storing whatever is returned by the database. When we have 'null' in the db the frontend is storing and returning null. Same for an empty object.

levity:
Hmm. Are you sure there are no conditions (e.g. a certain sequence of user
events in editing) on the frontend that would lead to the replacement of a
blank value with curly braces? For example, in the recent mobile profile
editing bug, where Sam and Ed had their names changed to other users’
names, they also had their bio and tag line fields changed to ‘{}’. So my
suspicion is that there is something with the frontend that causes this.

It’s possible that it’s already been fixed, too. In this case what I would
recommend is simply repairing any bad values in the database and waiting to
see if the problem ever reappears.

No need for a code change that seems speculative to me. Although you could
add Rollbar logging, so if the issue does reoccur, we’re proactively
notified about it.

PR #388, opened 2017-12-13
Return null appropriately

andygimma:


andygimma:
@levity the function convertGraphqlData is responsible for changes those null values to empty objects before saving to the database. This fix works with user settings, but I wonder if it will have unintended side-effects.

levity:
This looks good to me. The root issue here is JavaScript weirdness: `typeof null === ‘object’`

PR #390, opened 2017-12-18
Group memberships for posts (DEV-167)

levity:
This is the work spun off from #378 that pertains to replacing Follow and LastRead with GroupMembership between User and Post. It's ready to go and has a migration that will set up all the existing data in the new schema.

### QA

Visual verify on this should be done by deploying it to staging so that the migration runs there. Behavior around following and the last read timestamp -- i.e. starting to follow a post due to commenting, receiving notifications for comments on followed posts, viewing your list of threads, not getting notified about new messages for the the thread you're currently viewing, etc. -- should be unchanged.


api/graphql/makeModels.js
```diff
@@ -272,12 +272,7 @@ export default async function makeModels (userId, isAdmin) {
       relations: [
         {followers: {alias: 'participants'}},
         {comments: {alias: 'messages', typename: 'Message', querySet: true}}
-      ],
-      filter: nonAdminFilter(relation => relation.query(q => {
```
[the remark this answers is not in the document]
levity: Ah, good catch, I need to replace this. It limits you to message threads that you're following.

api/models/Comment.js
```diff
@@ -60,44 +55,34 @@ module.exports = bookshelf.Model.extend(Object.assign({
     return Promise.resolve([])
   },
 
-  createActivities: function (trx) {
-    const isReplyToPost = !this.get('comment_id')
-
-    var toLoad = ['post', 'post.followers']
-    if (!isReplyToPost) toLoad = toLoad.concat(['comment', 'comment.followers'])
-
```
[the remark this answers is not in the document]
levity: What I took out was the logic related to nested comments -- it was never used in production so I figured it would be cleaner to throw it out because if and when we want to use it again, it'll be done differently anyway with GroupConnection

api/graphql/makeModels.js
```diff
@@ -273,11 +274,9 @@ export default async function makeModels (userId, isAdmin) {
         {followers: {alias: 'participants'}},
         {comments: {alias: 'messages', typename: 'Message', querySet: true}}
       ],
-      filter: nonAdminFilter(relation => relation.query(q => {
+      filter: relation => relation.query(q =>
```
robbiecarlton: Why remove `nonAdminFilter` here?

api/models/Community.js
```diff
@@ -92,28 +93,22 @@ module.exports = bookshelf.Model.extend(merge({
 
   createStarterPosts: function (transacting) {
     var now = new Date()
-    var timeShift = {null: 0, intention: 1, offer: 2, request: 3}
-    return Community.find('starter-posts', {withRelated: [
-      'posts', 'posts.followers', 'posts.selectedTags'
-    ]})
+    var timeShift = {offer: 1, request: 2}
```
robbiecarlton: Looks like we only have 1 starter post in new communities now, so I think we can remove this timeshift stuff. And not in scope for this work so no worries if you don't

lib/groups-spike.js
```diff
@@ -0,0 +1,194 @@
+/* globals LastReadDeprecated, FollowDeprecated */
```
robbiecarlton: Is this file going to be left in the code? If so, is there a better name for it?

api/models/user/HasGroupMemberships.js
```diff
@@ -0,0 +1,27 @@
...
+      'group_memberships.active': true
+    })
+    .select('group_data_id')
+
+    if (where) subq = subq.where(where)
+
+    return model.collection().query(q => q.where('id', 'in', subq))
+  },
+
+  communitiesNotImplementedYet () {
```
robbiecarlton: if you're leaving this in, could you add a comment explaining what still needs to be done.

api/graphql/makeModels.js
levity: Basically because there's no need for you to see other people's message threads even if you're an admin.

lib/groups-spike.js
levity: OK, I can move it somewhere into api/models/group as a util or migration file.

levity:
🤟

PR #395, opened 2018-01-09
MBL-327 - Password reset for mobile 

lorenjohnson:


test/unit/controllers/SessionController.test.js
```diff
@@ -138,6 +139,19 @@ describe('SessionController', function () {
       })
     })
 
+    it('logs a user in and redirects (API/POST request)', () => {
```
robbiecarlton: I think this test is saying it "Doesn't" redirect

PR #396, opened 2018-01-11
Enable creating a community in a network WEB-381

robbiecarlton:


api/graphql/mutations/community.js
```diff
@@ -67,8 +67,14 @@ export function regenerateAccessCode (userId, communityId) {
   })
 }
 
-export function createCommunity (userId, data) {
-  return Community.create(userId, data)
```
razorman8669: It seems the old behavior was to return the community membership tuple, but it looks like you changed the behavior to only return the membership.  Is that going to break existing functionality?

robbiecarlton: pretty sure the old behavior was just to return the membership. See line 72 + 73 below

PR #399, opened 2018-01-31
Web 351: create topic

richchurcher:
As discussed, the last use of `tags` prior to its unlamented demise.

Note the changes to `Tag.find` may or may not need to be echoed in your topics rework, depending on whether or not you're finding topics by id. The old system guaranteed tags would never start with a number, but we are no longer using that regex so it's theoretically possible for someone to use a big number as a topic, which the old code tried to convert silently to a database id. Using named properties just makes it explicit.

This also upgrades Mocha, Chai, and the various Chai plugins to their latest versions.


api/models/Tag.js
```diff
@@ -146,23 +133,43 @@ module.exports = bookshelf.Model.extend({
...
+  find: function ({ id, name }, options) {
+    if (id) {
+      return Tag.where({ id }).fetch(options)
+    }
+    if (name) {
+      return Tag.query(qb => qb.whereRaw('lower(name) = lower(?)', name))
+        .fetch(options)
     }
-    return Tag.where({id: id}).fetch(options)
+    return Promise.resolve(null)
```
razorman8669: sure you wanna return resolve with null?  any repercussions?

richchurcher: That was the existing behaviour, so I didn't wanna mess with it. In this case I think it will simply go away when Robbie dumps the table :smile:

richchurcher:
@robbiecarlton this one's probably ready for a look :smile:

api/graphql/mutations/topic.js
```diff
@@ -1,4 +1,34 @@
-export function subscribe (userId, topicId, communityId, isSubscribing) {
-  return TagFollow.subscribe(topicId, userId, communityId, isSubscribing)
-  .then(() => ({success: true}))
+import { sanitize } from 'hylo-utils/text'
+
+export async function topicMutationPermissionCheck (userId, communityId) {
```
robbiecarlton: just a note that some mutations use `userId` here and others use an `authZ` object which is
```
{
  userId,
  isAdmin
}
```
Not saying you need to switch for this PR. But at some point we should standardize that.

api/graphql/mutations/topic.js
```diff
@@ -1,4 +1,34 @@
...
+
+export async function createTopic (userId, topicName, communityId) {
+  await topicMutationPermissionCheck(userId, communityId)
+  const name = sanitize(topicName)
+  const invalidReason = Tag.validate(name)
+  if (invalidReason) {
+    throw new Error(invalidReason)
+  }
+
+  const topic = await Tag.findOrCreate(name)
```
robbiecarlton: the variable naming makes me a little queasy. But I think the best change will be to rename Tag to Topic, which is obv outside of the scope of this ticket.
I was working on a branch that had that change, but have abandoned it as it was going down a wrong path.

api/models/Tag.js
```diff
@@ -145,23 +132,43 @@ module.exports = bookshelf.Model.extend({
...
   },
 
   tagsInText,
 
-  find: function (id, options) {
-    if (!id) return Promise.resolve(null)
-    if (isNaN(Number(id))) {
-      return Tag.query(qb => qb.whereRaw('lower(name) = lower(?)', id))
-      .fetch(options)
+  find: function ({ id, name }, options) {
```
robbiecarlton: I like this change

PR #405, opened 2018-02-21
fixes issue with findOrCreate thread

razorman8669:


razorman8669:
@robbiecarlton I ended up merging this since I'm pretty sure it fixes the problem.  but wanted you to take a look at it nonetheless.  let me know if that would break other edge cases?  basically, it's now searching for `id IN subquery` instead of `id = subquery` and THEN it's filtering out based on `posts.type === 'thread'`.  I had to do this since the subquery had no join on posts.

PR #406, opened 2018-03-06
force update for versions < 2.1

robbiecarlton:


robbiecarlton:
@richchurcher - for your consideration, the process for forcing old versions of the app to update

PR #409, opened 2018-03-06
Web 139 remove moderator

razorman8669:


api/graphql/mutations/community.js
```diff
@@ -12,9 +12,14 @@ export async function addModerator (userId, personId, communityId) {
   return community
 }
 
-export async function removeModerator (userId, personId, communityId) {
+export async function removeModerator (userId, personId, communityId, isRemoveFromCommunity) {
   const community = await getModeratedCommunity(userId, communityId)
-  await GroupMembership.removeModeratorRole(personId, community)
+  if (isRemoveFromCommunity) {
+    await CommunityService.removeMember(personId, communityId)
```
richchurcher: The only thing I wonder about here, and it's probably extra-picky, is that the moderator bit is effectively still set (even though the user is now `active = false` through being removed from the community). If the user is later re-added, oops, they're a moderator now! Would it be worth doing both, first zero the moderator role, then `removeMember`?

api/graphql/mutations/community.test.js
```diff
@@ -52,11 +52,26 @@ describe('moderation', () => {
   })
 
   describe('removeModerator', () => {
-    it('works', async () => {
+    it('just removes moderator role', async () => {
```
richchurcher: For extra credit, you could add one that checks to ensure the mutation can't be used by a non-moderator :smile:

api/graphql/mutations/community.js
richchurcher: Unless the user add code takes care of that, in which case we're fine. But I have a suspicion it might not.

razorman8669: nice catch.  I fixed as suggested (doing both)

PR #411, opened 2018-03-15
Web 22 community announcements

andygimma:


api/graphql/schema.graphql
```diff
@@ -233,9 +233,11 @@ type Post {
   attachments(type: String): [Attachment]
   attachmentsTotal: Int
   postMemberships: [PostMembership]
+  postMembershipsTotal: Int
```
robbiecarlton: duplication

knexfile.js
```diff
@@ -32,5 +32,14 @@ module.exports = {
   development: defaults,
   dummy: Object.assign({}, defaults, { seeds: { directory: './seeds/dummy' } }),
   staging: defaults,
-  production: merge({connection: {ssl: true}}, defaults)
+  production: merge({connection: {ssl: true}}, defaults),
+  docker: Object.assign({},
```
robbiecarlton: should this be part of this PR? Not sure what it's doing

migrations/20180315140159_add_announcement_to_post.js
```diff
@@ -0,0 +1,12 @@
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table('posts', t => {
+    t.boolean('announcement').defaultTo(false)
+  })
+};
+
+exports.down = function(knex, Promise) {
```
robbiecarlton: the down function should drop the column ie
```
  return knex.schema.table('posts', t => t.dropColumn('announcement'))

```

migrations/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+postgres:
```
robbiecarlton: same question

knexfile.js
andygimma: @richchurcher made this change to allow for easier migrations.

robbiecarlton: cool

PR #414, opened 2018-03-28
check update for mobile uses env variable DEV-301

robbiecarlton:
When merging this be sure to add MINIMUM_SUPPORTED_MOBILE_VERSION to heroku env variables set to the latest mobile version

### DIFF
--- a/api/controllers/MobileAppController.js
+++ b/api/controllers/MobileAppController.js
@@ -11,6 +11,7 @@ module.exports = {
     }
     res.ok(result)
   },
+
   checkShouldUpdate: function (req, res) {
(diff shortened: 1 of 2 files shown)


robbiecarlton:
interesting question. For sure we could do it, but it seems unlikely that it will happen in practice. Trying to imagine a scenario where an apk or ipa is behind the production build version number

lorenjohnson:
I think it can happen fairly easily from a feature branch atm due to how fragile our version # pipeline is… It’s fine to just wait to see if it comes-up. If it does it can be fixed by re-building after merging from master.

PR #417, opened 2018-04-10
Fix versioncheck

richchurcher:
This is a pretty trivial one. Currently the request will return `undefined` for the most part (if a user has the correct version). Even though the status code is correct, the fact that we're returning an empty body for an `application/json` request is technically an error because empty string is not, as it turns out, valid JSON :laughing: so there has been a swallowed error on startup for a long time.

Once I started logging it it took me quite a while to realise that it wasn't the source of our problems with the dev environment, but we should still fix it. Swallowed errors ain't cool.

Note: the client code ignores the body unless it has an error property, so this doesn't break anything.

### DIFF
--- a/api/controllers/MobileAppController.js
+++ b/api/controllers/MobileAppController.js
@@ -17,7 +17,6 @@ module.exports = {
     const androidVersion = req.param('android-version')
     const version = iosVersion || androidVersion
     const platform = iosVersion ? IOS : ANDROID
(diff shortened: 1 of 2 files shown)


robbiecarlton:
oh, except a test is failing. Fix that and you're good to go

richchurcher:
Empty strings should be valid EVERYTHING. Except maybe numbers. And objects. And arrays. And null. Hrm...

PR #422, opened 2018-04-23
MBL-481 Add push notification case

andygimma:


api/models/Notification.js
```diff
@@ -54,6 +54,7 @@ module.exports = bookshelf.Model.extend({
     var action
     switch (this.get('medium')) {
       case MEDIUM.Push:
+        // TODO check here
```
robbiecarlton: remove this comment

api/models/Notification.js
```diff
@@ -89,11 +90,25 @@ module.exports = bookshelf.Model.extend({
...
       case 'approvedJoinRequest':
         return this.sendApprovedJoinRequestPush()
+      case 'announcement':
+        return this.sendPushAnnouncement()
       default:
         return Promise.resolve()
     }
   },
 
+  sendPushAnnouncement: function (version) {
```
robbiecarlton: you could test this by using mockify* from test/setup/helpers to replace this.reader().sendPushNotification with a mock function and then see if it was called with the expected alertText and path

*doesn't look like there are any examples of using mockify with instance methods, but I don't see why it shouldn't work just as well as with class methods. If mockify doesn't work you should be able to replace it with a mock by hand so to speak.

api/models/Notification.js
```diff
@@ -89,11 +90,25 @@ module.exports = bookshelf.Model.extend({
...
+    var post = this.post()
+    var communityIds = Activity.communityIds(this.relations.activity)
+    if (isEmpty(communityIds)) throw new Error('no community ids in activity')
+    return Community.find(communityIds[0])
+    .then(community => {
+      var path = url.parse(Frontend.Route.post(post, community)).path
+      var alertText = PushNotification.textForPost(post, community, this.relations.activity.get('reader_id'), version)
```
robbiecarlton: Do you want the announcement notification text to be different in any way from a normal post? Probably. Even if it's just appending "Announcement: " to the front. If so you will need to either expand textForPost or create a different method (PushNotification.textForAnnouncement) to be used here.

PR #426, opened 2018-05-17
Find and send invitation

andygimma:


api/models/Invitation.js
```diff
@@ -80,6 +80,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
   },
 
   send: function () {
+    console.log('in send *****')
```
razorman8669: Remove console statements

PR #441, opened 2018-09-04
Fix invitation moderator override condition, adding a test would be good

lorenjohnson:


api/services/InvitationService.js
```diff
@@ -167,26 +167,19 @@ module.exports = {
...
-          // preexisting = true
-          return true
-        } else {
+      .then(community => {
+        return GroupMembership.forPair(user, community, {includeInactive: true}).fetch()
+        .then(existingMembership => {
+          if (existingMembership) return existingMembership.get('active')
+              ? existingMembership
+              : existingMembership.save({active: true}, {patch: true}).then(membership => membership)
+          if (!!community) return user.joinCommunity(community).then(membership => membership)
```
robbiecarlton: tiny quibble, I would put this check at the top of the block, like
`if (!community) return` or `if (!community) throw ...`
think it reads clearer there. But feel free not to change.

PR #447, opened 2018-12-21
Web-574 - Crowd funding

robbiecarlton:


api/graphql/makeModels.js
```diff
@@ -140,7 +142,7 @@ export default async function makeModels (userId, isAdmin) {
         commentsTotal: p => p.get('num_comments'),
         votesTotal: p => p.get('num_votes'),
         type: p => p.getType(),
-        myVote: p => p.userVote(userId).then(v => !!v)
+        myVote: p => p.userVote(userId).then(v => !!v),
```
lorenjohnson: trailing comma. no big deal

api/graphql/mutations/project.js
```diff
@@ -1,5 +1,6 @@
 import { createPost } from './post'
 import { uniq } from 'lodash/fp'
+var stripe = require("stripe")(process.env.STRIPE_API_KEY);
```
lorenjohnson: We're going to push this through today and discuss if it proves relevant later, but I think I would prefer this stuff kept in it's own mixin and closer to the model not the API. Like `models/post/project/contributions.js` or `models/project/contributions.js` or if it made any sense to generalize more `models/contributions`... For later, but let's keep it in mind because if that makes sense it'd be nice to give this thought before embedding more at this layer.

PR #449, opened 2019-03-07
DEV-406 - Topics Everywhere

lorenjohnson:


tibetsprague:
Also adds default/suggested and pinned topics https://github.com/Hylozoic/hylo-node/issues/468

PR #456, opened 2019-05-03
DEV-431 comment notification bug

robbiecarlton:
this one is really weird

### DIFF
--- a/lib/graphql-bookshelf-bridge/fetcher.js
+++ b/lib/graphql-bookshelf-bridge/fetcher.js
@@ -27,11 +27,14 @@ export default class Fetcher {
       type = 'hasMany' // n.b. we're not supporting belongsTo yet
     }
 
-    if (type === 'belongsTo' && !parentFk) return Promise.resolve()
-
     if (!typename) typename = this._getTypenameFromTableName(targetTableName)
 
     const loader = this.loaders[typename]
+
+    if (type === 'belongsTo') {
+      if (!parentFk) return Promise.resolve()
+      return loader.load(parentFk)
+    }
 
     // apply the filter that always applies to this target model, if any
     const model = this._getModel(typename)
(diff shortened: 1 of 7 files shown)


lib/graphql-bookshelf-bridge/fetcher.js
```diff
@@ -27,11 +27,15 @@ export default class Fetcher {
       type = 'hasMany' // n.b. we're not supporting belongsTo yet
     }
 
-    if (type === 'belongsTo' && !parentFk) return Promise.resolve()
-
     if (!typename) typename = this._getTypenameFromTableName(targetTableName)
+
     const loader = this.loaders[typename]
 
+    if (type === 'belongsTo') {
```
robbiecarlton: this is the change that fixed things. I just reverted to older code. I think the change was yours, (it was part of the blocked user work), so if you can remember why you made the change it would be helpful to see if this has broken something else.
All tests are passing, I haven't done any QA to see if I can figure out what that change was for, will do some in the morning
It's a little unsatisfying (as I suspected it might be).
Honestly don't totally understand why things broke the way they did because of this change but this definitley fixes it. I might poke around some more in the morning to try and understand it, but also might just move on

lorenjohnson: I definitely made the change and I would have to dig back into it a bit to understand why. Look here at the code before my change (Lines 33-36):

https://github.com/Hylozoic/hylo-node/blame/a918f0ce4e690bdd158a80bd7ec82a6f99a518a1/lib/graphql-bookshelf-bridge/fetcher.js

lorenjohnson: So long and short, it was related to getting the blocked users query filtering working. If you QA that feature I am going to guess that something breaks, and frankly that's better then notifications breaking--I guess. But of course we want both to work.

PR #480, opened 2020-05-21
WIP public posts

### LINKED ISSUE #466: Handle Public Posts on the backend
Add field to determine whether a post should be public to everyone or only visible to the communities it is shared to.

- public/private flag - add `public` column to posts table which takes a boolean
- We decided to wait on this and just do a boolean for now: Create a `Public` community that public posts are added to by creating a record in posts_communities join table connecting the post to the public community

cpolitano:
[closes #466 ]

- very in progress
- so far just a migration to add a `public` column to posts table that takes a boolean

### DIFF
--- a/migrations/20200513172754_add-public-posts.js
+++ b/migrations/20200513172754_add-public-posts.js
@@ -1,0 +1,12 @@
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table('posts', table => {
+    table.boolean('public').defaultTo(false)
+  })
+};
+
+exports.down = function(knex, Promise) {
+  return knex.schema.table('posts', table => {
+    table.dropColumn('public')
+  })
+};


api/graphql/makeModels.js
```diff
@@ -133,13 +133,14 @@ export default async function makeModels (userId, isAdmin) {
...
         'announcement',
-        'accept_contributions'
+        'accept_contributions',
+        'is_public'
       ],
       getters: {
         title: p => p.get('name'),
         details: p => p.get('description'),
         detailsText: p => p.getDetailsText(),
-        public: p => (p.get('visibility') === Post.Visibility.PUBLIC_READABLE) || null,
```
tibetsprague, line 142: whoa, what was this?? im intrigued. looks like they were already working on this. did you think about just using what they already had? looks like they were planning for a few different visibility options.

api/graphql/makeModels.js
```diff
@@ -191,7 +192,7 @@ export default async function makeModels (userId, isAdmin) {
         'created_at',
         'avatar_url',
         'banner_url',
-        'num_members',
+        'memberCount',
```
tibetsprague, line 195: I'm a bit confused why most attributes use underscores but postCount is is not. generally would prefer to stay consistent, but also not a big deal. 
also curious why you couldn't use the existing num_members? i do like having memberCount match the naming of postCount

api/graphql/schema.graphql
```diff
@@ -139,7 +139,6 @@ type Community {
   settings: String
   invitePath: String
   network: Network
-  numMembers: Int
```
tibetsprague, line 142: looking at the database it seem that num_members is being set and is correct, so why not just use that?

api/models/Community.js
```diff
@@ -141,6 +141,11 @@ module.exports = bookshelf.Model.extend(merge({
   },
 
   memberCount: function () {
+    // return this.get('num_members')
+    return this.users().fetch().then(x => x.length)
+  },
+
+  numMembers: function () {
```
tibetsprague, line 148: seems like we could just use this cached value that already exists instead of calculating it every time

api/models/Post.js
```diff
@@ -176,7 +176,8 @@ module.exports = bookshelf.Model.extend(Object.assign({
   },
 
   isPublic: function () {
-    return this.get('visibility') === Post.Visibility.PUBLIC_READABLE
+    // return this.get('visibility') === Post.Visibility.PUBLIC_READABLE
```
tibetsprague, line 179: lets decide which to use and remove comment if we want to go with the boolean

api/graphql/makeModels.js
cpolitano: Solution - going with the boolean `public` column for now, will investigate Visibility option for the future and we solidify the architecture of public posts & communities.

api/graphql/makeModels.js
cpolitano: as discussed, the camel case is composed attributes and the snake case comes directly from the db

api/models/Post.js
cpolitano: 👍

api/models/Community.js
cpolitano: please see the function on line 203-208, `reconcileNumMembers`. After noticing the member counts from `num_members` were off by 1 for me locally, I noticed this function which tipped me off that num_members isn't necessarily reliable. We agreed to investigate this later, not resolve it now.

api/models/flaggedItem/notifyUtils.js
```diff
@@ -4,25 +4,27 @@ export async function notifyModeratorsPost (flaggedItem) {
...
+  const isPublic = post.attributes.is_public
+  return sendToCommunities(flaggedItem, communities, isPublic)
 }
 
 export async function notifyModeratorsMember (flaggedItem) {
   const member = await flaggedItem.getObject()
   const user = flaggedItem.relations.user
   const communities = await user.communitiesSharedWithUser(member)
-  return sendToCommunities(flaggedItem, communities)
+  return sendToCommunities(flaggedItem, communities, isPublic = false)
```
tibetsprague, line 24: just pass in false, no need to set a variable. im suprised there's no lint complaint about this

api/models/flaggedItem/notifyUtils.js
```diff
@@ -33,8 +35,10 @@ export async function sendToCommunities (flaggedItem, communities) {
     await send(community, moderators.map(x => x.id))
   }
 
-  const shouldSendToAdmins = process.env.HYLO_ADMINS &&
```
tibetsprague, line 36: whoa, this is something that already existed?

tibetsprague:
just realized we probably want an index on the is_public column too, thats a quick and easy migration

PR #483, opened 2020-05-21
467-resources

### LINKED ISSUE #467: New post type "resource" back-end
This post represents a physical resource. e.g. A lemon tree that people can pick from, or a community garden plot with open beds…

- [x] Location is required 
- [x] Track if resource is still available or not
- [x] Include time-frame (end times) to availability of resources

- [ ] Question: do we want to track amounts as a separate field? Like there are 5 apple trees to give away here

- [ ] Question: How is this different from an Offer? Maybe it’s just a specific type of offer? But it does have a different feel, more long running. More meant for the map. It could in the future have additional fields like amount, cost, etc. and eventually could lead to a marketplace.

cpolitano:
[Closes #467 ]

### DIFF
--- a/api/models/Post.js
+++ b/api/models/Post.js
@@ -158,7 +158,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
       var type = this.get('type')
       if (type) return type
       const tagNames = this.relations.tags.map(t => t.get('name'))
-      const typeNames = intersection(tagNames, ['request', 'offer'])
+      const typeNames = intersection(tagNames, ['request', 'offer', 'resource'])
       if (!isEmpty(typeNames)) {
         return typeNames[0]
       } else {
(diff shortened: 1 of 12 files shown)


api/models/Post.js
```diff
@@ -491,10 +492,10 @@ module.exports = bookshelf.Model.extend(Object.assign({
 
   fixTypedPosts: () =>
     bookshelf.transaction(transacting =>
-      Tag.where('name', 'in', ['request', 'offer', 'intention'])
+      Tag.where('name', 'in', ['request', 'offer', 'resource', 'intention'])
```
tibetsprague, line 495: huh, I wonder what intention is/was? i also wonder if this is old code using Tags for type instead of the post type column

api/models/post/validatePostData.js
```diff
@@ -4,16 +4,23 @@ export default function validatePostData (userId, data) {
...
   if (isEmpty(data.community_ids)) {
     throw new Error('no communities specified')
   }
+
   if (data.topicNames && data.topicNames.length > 3) {
     throw new Error('too many topics in post, maximum 3')
   }
+  
+  if (data.type === Post.Type.RESOURCE && !data.location_id) {
+    throw new Error('resources must have a location')
```
tibetsprague, line 22: I've actually been wondering recently if we might not want to use resources for other types of resources like books, links, files... that the community wants to kind of "store" or track... but thats for the future. location based ones only for now is good!

PR #493, opened 2020-06-07
Await for relation to load before calling community in activity

lorenjohnson:
Fixes issue with creating new comments. The `Post.communities` relationship was not loaded in the new Comment context. Now forced loaded and no longer breaks.

### DIFF
--- a/api/models/Comment.js
+++ b/api/models/Comment.js
@@ -26,7 +26,8 @@ module.exports = bookshelf.Model.extend(Object.assign({
     return this.hasMany(Thank)
   },
 
-  community: function () {
+  community: async function () {
+    await this.relations.post.load(['communities'])
     return this.relations.post.relations.communities.first()
   },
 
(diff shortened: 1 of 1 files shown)


api/models/Comment.js
```diff
@@ -26,7 +26,8 @@ module.exports = bookshelf.Model.extend(Object.assign({
     return this.hasMany(Thank)
   },
 
-  community: function () {
+  community: async function () {
```
tibetsprague, line 29: could there be any weird side effects of making this call async? did you look around the code in the other places it is used?

lorenjohnson: I had, yes. It appears, and clicks for me that, that Comment#community is actually not used anywhere else, but bookshelf is tricky and I couldn't guarantee it. Finger tested things. If we have a staging backend/frontend going it'd be a good thing, but the testing I did against the current Production snapshot came-up clean.

tibetsprague: Cool. Staging is up and running so I can test there

PR #497, opened 2020-06-18
488 public communities

### LINKED ISSUE #488: Create Public Community options
Backend of Community Privacy Settings
- allow the public to see community and request to join
- anyone can join community
- anyone can see who is a member of community

These settings will be reflected in new db columns in `communities` table: `is_public` `is_auto_joinable` and `public_member_directory` which are all booleans

Depends on [#466 ]

cpolitano:
[closes #488 ]

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -198,7 +198,10 @@ export default async function makeModels (userId, isAdmin) {
         'postCount',
         'location',
         'hidden',
-        'allow_community_invites'
+        'allow_community_invites',
+        'is_public',
+        'is_auto_joinable',
+        'public_member_directory'
       ],
       relations: [
         'locationObject',
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -247,20 +247,23 @@ export default async function makeModels (userId, isAdmin) {
       ],
       getters: {
         feedItems: (c, args) => c.feedItems(args),
+        isPublic: c => c.get('is_public'),
         pendingInvitations: (c, { first }) => InvitationService.find({communityId: c.id, pendingOnly: true}),
         invitePath: c =>
           GroupMembership.hasModeratorRole(userId, c)
           .then(isModerator => isModerator ? Frontend.Route.invitePath(c) : null)
       },
       filter: nonAdminFilter(sharedNetworkMembership('communities', userId)),
-      fetchMany: ({ first, order, sortBy, offset, search, autocomplete, filter }) =>
+      fetchMany: ({ first, order, sortBy, offset, search, autocomplete, filter, isPublic, boundingBox, }) =>
         searchQuerySet('communities', {
+          boundingBox,
           term: search,
           limit: first,
           offset,
           type: filter,
           autocomplete,
-          sort: sortBy
+          sort: sortBy,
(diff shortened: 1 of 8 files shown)


api/services/Search.js
```diff
@@ -37,6 +37,10 @@ module.exports = {
         qb.orderBy(opts.sort || '', opts.order || 'asc')
       }
 
+      if (opts.is_public) {
+        qb.whereIn('is_public', opts.is_public)
```
tibetsprague, line 41: i dont think whereIn is correct, that is meant for array values, should just be where. but i guess it works!

migrations/20200617090839_add_community_privacy_settings.js
```diff
@@ -0,0 +1,24 @@
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table('communities', table => {
+    table.boolean('is_public').defaultTo(false)
+  }).then(() => knex.schema.table('communities', table => {
+    table.boolean('is_auto_joinable').defaultTo(false)
+  })).then(() => knex.schema.table('communities', table => {
+    table.boolean('public_member_directory').defaultTo(false)
+  })).then(() => knex.raw(`CREATE INDEX public_communities_idx ON communities USING btree (is_public);`))
```
tibetsprague, line 9: been reading up on indexes, i think we might want a single index that includes all three of these columns? but obviously not that important right now

api/graphql/makeModels.js
```diff
@@ -244,20 +247,23 @@ export default async function makeModels (userId, isAdmin) {
       ],
       getters: {
         feedItems: (c, args) => c.feedItems(args),
+        isPublic: c => c.get('is_public'),
```
tibetsprague, line 250: why do you need this, shouldn't all regular columns automatically be available to access?

api/services/Search.js
```diff
@@ -37,6 +37,10 @@ module.exports = {
...
+      }
+
```
tibetsprague, line 43: looks like you still need to add the boundingBox part right?

api/services/Search.js
cpolitano: we also used `whereIn` for public posts. I'll try changing both of them to `where`

migrations/20200617090839_add_community_privacy_settings.js
cpolitano: ok done 👍

api/services/Search.js
cpolitano: added ✔️

api/services/Search.js
```diff
@@ -38,9 +39,14 @@ module.exports = {
       }
 
       if (opts.is_public) {
-        qb.whereIn('is_public', opts.is_public)
+        qb.where('is_public', opts.is_public)
       }
 
+      filterAndSortCommunities({
+        search: opts.term,
+        sortBy: opts.sort,
```
tibetsprague, line 47: this seems like it is duplicating above filters, just a little worried it might cause issues, sorting one way and then another, or searching by something extra. maybe just take the search and sortBy out?

tibetsprague: in fact why isnt this code just using filterAndSortByCommunities for all of it?

Issue #499, opened 2020-06-22
Community join requests


tibetsprague:
- [x] Use existing join_requests table
- [x] Add a status column - an Integer. 0 = pending, 1 = accepted, 2 = rejected. Add an index on this column. Index should be multi column on community_id
- [x] existing updated_at will be used to track when status was set
- [x] delete existing join_request rows in the migration, or set them all to rejected
- [x] When someone is accepted add them to the community
- [x] Send notification when someone is accepted

PR #509, opened 2020-07-08
Community join requests

wushenshen:
addresses #499

### DIFF
--- a/api/graphql/mutations/join_request.js
+++ b/api/graphql/mutations/join_request.js
@@ -21,8 +21,8 @@ export async function createJoinRequest (communityId, userId) {
 }
 
 export async function acceptJoinRequest (joinRequestId, communityId, userId, moderatorId) {
-  await JoinRequest.update(joinRequestId, { status: 1 }, moderatorId)
   await joinCommunity(communityId, userId)
+  await JoinRequest.update(joinRequestId, { status: 1 }, moderatorId)
   return await JoinRequest.find(joinRequestId)
 }
 
--- a/api/graphql/mutations/join_request.js
+++ b/api/graphql/mutations/join_request.js
@@ -1,12 +1,13 @@
 export async function joinCommunity (communityId, userId) {
   const user = await User.find(userId)
   const community = await Community.find(communityId)
+
   return GroupMembership.forPair(user, community, {includeInactive: true}).fetch()
     .then(existingMembership => {
       if (existingMembership) return existingMembership.get('active')
         ? existingMembership
-        : existingMembership.save({active: true}, {patch: true}).then(membership => { console.log('membership exists ==>', membership); return membership })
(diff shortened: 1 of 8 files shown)


wushenshen:
@tibetsprague This addresses items 1, 2, and 4 from the issue - I chose to set all the existing rows to rejected (status = 2) rather than delete them, feels safer. Thoughts on this migration?

migrations/20200707214505_add_join_requests_status.js
```diff
@@ -0,0 +1,21 @@
...
+    table.index(['community_id', 'status'])
+  })
+
+  const now = new Date().toISOString()
+  await knex.raw(`UPDATE "join_requests" set status = 2, updated_at = '${now}'`)
+};
+  
+exports.down = async function(knex, Promise) {
+  await knex.schema.table('join_requests', table => {
+    table.dropIndex(['community_id', 'status'])
```
tibetsprague, line 14: dropping the column will automatically drop the index I believe but this is also fine

api/graphql/mutations/join_request.js
```diff
@@ -0,0 +1,32 @@
+export async function joinCommunity (communityId, userId) {
+  const user = await User.find(userId)
+  const community = await Community.find(communityId)
+
+  return GroupMembership.forPair(user, community, {includeInactive: true}).fetch()
+    .then(existingMembership => {
+      if (existingMembership) return existingMembership.get('active')
+        ? existingMembership
+        : existingMembership.save({active: true}, {patch: true}).then(membership => membership)
+      if (!!community) return user.joinCommunity(community).then(membership => membership)
```
tibetsprague, line 10: kind of seems like this check for the community should happen sooner in this function? not a big deal but why do anything if you cant find the user or the community?

api/graphql/mutations/join_request.js
```diff
@@ -0,0 +1,32 @@
...
+
+export async function createJoinRequest (communityId, userId) {
+  return JoinRequest.create({
+    userId: userId,
+    communityId,
+  })
+  .then(request => ({ request }))
+}
+
+export async function acceptJoinRequest (joinRequestId, communityId, userId, moderatorId) {
```
tibetsprague, line 23: do we need to check somewhere that this is being called by the userId only?

api/graphql/mutations/join_request.js
wushenshen: @tibetsprague good point - I pulled this line out and removed the rest - `user.JoinCommunity` leads to `group.addMembers` which covers the case of an existing membership

api/graphql/mutations/join_request.js
wushenshen: I added a couple checks for both user and community in the joinCommunity mutation, is that what you had in mind?

PR #519, opened 2020-07-31
Comment attachments

lorenjohnson:


api/models/Media.js
```diff
@@ -144,3 +137,22 @@ const createAndAddSize = function (attrs) {
 
   return Media.create(attrs)
 }
+
+export function makeSubjectIdKeyAndValue ({subjectId, subjectType, postId, commentId}) {
+  let subjectIdKey, subjectIdValue
+
+  if (subjectId && subjectType) {
```
tibetsprague, line 144: im not quite clear why you need subjectId and subjectType alongside postId and commentId, cant you just use subjectId and subjectType for everything?

api/models/post/createPost.js
```diff
@@ -33,30 +33,30 @@ export function afterCreatingPost (post, opts) {
 
     // Add media, if any
     // redux version
-    opts.imageUrl && Media.createForPost({
+    opts.imageUrl && Media.createForSubject({
       postId: post.id, type: 'image', url: opts.imageUrl
```
tibetsprague, line 37: could just use subjectId and subjectType here right?

api/models/post/createPost.js
```diff
@@ -33,30 +33,30 @@ export function afterCreatingPost (post, opts) {
...
     opts.fileUrls && Promise.map(opts.fileUrls, (url, i) =>
-      Media.createForPost({
+      Media.createForSubject({
         postId: post.id, type: 'file', url, position: i
       }, trx)),
 
     opts.children && updateChildren(post, opts.children, trx),
 
-    opts.docs && Promise.map(opts.docs, doc => Media.createDoc(post.id, doc, trx)),
+    // google doc / video not currently used in evo
```
tibetsprague, line 54: we do really want video support though

lib/uploader/index.js
```diff
@@ -29,7 +29,20 @@ export function upload (args) {
       converter.on('error', err => reject(err))
 
       storage = createS3StorageStream(type, id, {userId, fileType, filename})
-      storage.on('finish', () => resolve(storage.url))
+      storage.on('finish', () => {
+        let attachmentType = fileType && fileType.mime.split('/')[0]
+
+        attachmentType = ['file', 'image'].includes(attachmentType)
```
tibetsprague, line 35: sort of an odd way to add a default

api/models/post/createPost.js
lorenjohnson: Well yeah, but that's a whole other thing requiring a transcoding service and such.

api/models/post/createPost.js
lorenjohnson: Yes, I introduced that pattern as an option and used it throughout in Comment, but left support for this legacy way of doing things intact for Post as it's relied upon at a couple of layers. I guess I could fully refactor to this pattern throughout.

lib/uploader/index.js
lorenjohnson: Hmm. I'm curious what implementation you'd find less odd that'd achieve the logic required logic here and otherwise better express it? 

---
That was the TL;DR. Here is the long answer:

Lines 33-37 are expressing preeminently a transformation from mimetype to a valid Media.type which reverts to a default Media.type on the final line. I suspect that the oddness you're sensing here is symptomatic of the way Media.type is implemented as a meta category over mimetype more than the way I've implemented the logic, but maybe you're pointing to something more rudimentary. Either way I would say that the valid Media.types and it's default should probably be their own constants (VALID_MEDIA_TYPES and DEFAULT_MEDIA_TYPE, respectively) and in the best world those constants should be shared between the front and backend as they function as related delineations for both backend conversations as well as display logic. I think we're within relatively easy reach from this branch forward to migrating Media.type to Media.mimetype and this should all get a freshen up along with that...

api/models/post/createPost.js
tibetsprague: seems like it should be easy to do, if so do it, if its a big pain, no worries about it

api/models/post/createPost.js
tibetsprague: yes of course

lib/uploader/index.js
tibetsprague: nevermind, it was a smaller thing but actually I think this is fine

lorenjohnson:
@tibetsprague I've taking things to the next couple of logical steps in terms of the refactorings here, but haven't fully tested those updates yet to the extent I have what is here at this point. My recommendation is that, once approved here and on the related frontend brach that we merge and release and then review and QA this new PR as we have time and then release it. It is better / more clear for sure and would be nice to have in there.

PR #523, opened 2020-08-10
521 import posts from a CSV

tibetsprague:
one thing that would be nice to add soon is real time progress indicator on the front-end showing how many posts have been imported and which ones have errors and aren't imported, https://github.com/Hylozoic/hylo-evo/issues/599


tibetsprague:
im on it, almost done

PR #534, opened 2020-09-12
Stop looking up a post's type in its tags

tibetsprague:
We don't need to do that anymore. This eliminates many many queries!


cpolitano (approved):
LGTM

PR #542, opened 2020-10-12
Fixes #538 - Recent Activity query is broken

### LINKED ISSUE #538: Fetching recent activity on member profile is broken
It is joining posts table twice

lorenjohnson:


lorenjohnson:
There was an extra, unnecessary, join to the Posts table in the Comments relation. This means that Recent Activity would have been ALWAYS broken unless there was already a cached query with the data in the front-end from other queries. This fixes that.

wushenshen (approved):

api/models/User.js
```diff
@@ -17,7 +17,6 @@ module.exports = bookshelf.Model.extend(merge({
   comments: function () {
     return this.hasMany(Comment)
     .query(q => {
-      q.join('posts', 'posts.id', 'comments.post_id')
```
tibetsprague, line 20: and we are sure this doesnt break comments anywhere else?

lorenjohnson: I am as sure as I can be in having tested through the UI comment creation and presentation on Posts. No issues arose there. I wouldn't expect any trouble, but I'm confused why this extra and self-apparently redundant `join` was added in the first place. I have to for now assume it was added in error.

lorenjohnson:
@tibetsprague this is ready for merge anytime.

tibetsprague (approved):

PR #545, opened 2020-10-15
514 saved searches

### LINKED ISSUE #514: Saved searches back-end
new table saved_searches:
- id
- user_id
- name (optional string)
- context (community, network, all, or public)
- context_id (optional): either a community_id or network_id or empty for all or public
- search_text
- post_types (array)
- bounding_box
- last_post_id (which was the last post that was found in this search and already sent to the user)

also table saved_search_topics (to track which topics to include in the search):
- id
- tag_id
- saved_search_id

We then need a worker on the back-end that once a day (maybe we want this to happen more often but lets start with this) checks all saved searches and sends notifications to the user for any new posts that match the search. 

---

- [x] Add query for fetching all saved searches
- [x] Add mutation for creating a saved search
  - [x] Include validation for context
  - [x] Include creating saved search topics
  - [x] Include saving last post id
- [x] Add mutation for deleting a saved search (by marking it inactive)
- [x] Worker job
   - [x] Send a digest for each saved search according to the user's existing preference for weekly or daily digests
   - [ ] ~~Include a section in the existing daily or weekly digest for saved searches. Show each search separately so the same post could appear multiple times~~
   - [ ] ~~Also send push notifications and in-app notifications if there are any new posts that match any of your searches. "There are new posts in this saved search (if only one new post link to that specific post in that map view, if multiple just link to the saves search)~~

wushenshen:
Closes #514 

This includes the migrations, schema changes, and plumbing to implement saved searches and saved search topics. It also adds a module to `lib/community/digest2` for the saved search digest, see below for an example.

Note that this only sends notifications by email, it does not send in-app notifications, that would be a future enhancement. Also, clicking the "View on Hylo" button takes the user to the post view on the map, but it's not centered on the post yet. Could bring it to the post in the stream view instead.

Screen Shot 2020-10-14 at 8 35 04 PM

Related to https://github.com/Hylozoic/hylo-evo/pull/663


lorenjohnson (approved):
👍  LGTM! 

* Adding some test coverage for this feature here would be good, but the hylo-node test suite is behind as it is so it can be difficult to navigate...

tibetsprague (approved):

PR #555, opened 2020-11-02
Add query for a member's projects

### LINKED ISSUE #552: Query projects in a given context
Needed for https://github.com/Hylozoic/hylo-evo/issues/631 - query all projects a user has created or joined and return only the ones accessible to the logged in user.

When we implement landing pages we will also need to query projects within a community, network, or other group.

wushenshen:
Closes #552

### DIFF
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -165,12 +165,7 @@ export function makeAuthenticatedQueries (userId, fetchOne, fetchMany) {
     post: (root, { id }) => fetchOne('Post', id),
     posts: (root, args) => fetchMany('Post', args),
     people: (root, args) => fetchMany('Person', args),
-    projects: (root, { context, contextId, viewerId }) => {
-      return Post.projectsForContext(context, contextId)
-        .then(({ models }) => {
-          return { items: models.filter(async p => {return await p.isVisibleToUser(p.id, viewerId) })}
-        })
-    },
+    projects: async (root, { context, contextId, viewerId }) => Post.projectsForContext(context, contextId, viewerId),
     connections: (root, args) => fetchMany('PersonConnection', args),
     communityTopics: (root, args) => fetchMany('CommunityTopic', args),
     topics: (root, args) => fetchMany('Topic', args),
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -165,6 +165,12 @@ export function makeAuthenticatedQueries (userId, fetchOne, fetchMany) {
     post: (root, { id }) => fetchOne('Post', id),
     posts: (root, args) => fetchMany('Post', args),
     people: (root, args) => fetchMany('Person', args),
+    projects: (root, { context, contextId, viewerId }) => {
+      return Post.projectsForContext(context, contextId)
+        .then(({ models }) => {
+          return { items: models.filter(async p => {return await p.isVisibleToUser(p.id, viewerId) })}
+        })
+    },
     connections: (root, args) => fetchMany('PersonConnection', args),
     communityTopics: (root, args) => fetchMany('CommunityTopic', args),
     topics: (root, args) => fetchMany('Topic', args),
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -165,7 +165,6 @@ export function makeAuthenticatedQueries (userId, fetchOne, fetchMany) {
     post: (root, { id }) => fetchOne('Post', id),
     posts: (root, args) => fetchMany('Post', args),
     people: (root, args) => fetchMany('Person', args),
-    projects: async (root, { context, contextId, viewerId }) => Post.projectsForContext(context, contextId, viewerId),
     connections: (root, args) => fetchMany('PersonConnection', args),
     communityTopics: (root, args) => fetchMany('CommunityTopic', args),
     topics: (root, args) => fetchMany('Topic', args),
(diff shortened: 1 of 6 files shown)


api/models/Post.js
```diff
@@ -362,6 +362,26 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+        postIds = await this.query(q => {
+          q.select('posts.*')
+          q.from('group_memberships')
+          q.leftJoin('groups', 'group_memberships.group_id', 'groups.id')
+          q.leftJoin('posts', 'groups.group_data_id', 'posts.id')
+          q.whereNotNull('group_memberships.project_role_id')
+          q.andWhere('group_memberships.user_id', contextId)
+          q.andWhere('group_memberships.active', true)
+        }).query()
+          .then(rows => rows.map(r => r.id))
```
tibetsprague, line 379: why are you doing multiple queries, first getting ids then loading those ids again?

api/models/Post.js
```diff
@@ -362,6 +362,26 @@ module.exports = bookshelf.Model.extend(Object.assign({
     PUBLIC_READABLE: 1
   },
 
+ projectsForContext: async function (context, contextId) {
+    let postIds
+    
+    switch (context) {
+      default: // For now, the only context is member profiles. Add to this once we show projects on group landing pages.
+        postIds = await this.query(q => {
+          q.select('posts.*')
```
tibetsprague, line 371: if you are only using the ids from this query then you can just select posts.id

api/graphql/index.js
```diff
@@ -165,6 +165,12 @@ export function makeAuthenticatedQueries (userId, fetchOne, fetchMany) {
     post: (root, { id }) => fetchOne('Post', id),
     posts: (root, args) => fetchMany('Post', args),
     people: (root, args) => fetchMany('Person', args),
+    projects: (root, { context, contextId, viewerId }) => {
+      return Post.projectsForContext(context, contextId)
+        .then(({ models }) => {
+          return { items: models.filter(async p => {return await p.isVisibleToUser(p.id, viewerId) })}
```
tibetsprague, line 171: we dont otherwise do this type of filtering here, seems like this should happen in the projectsForContext function maybe?

api/graphql/index.js
```diff
@@ -165,6 +165,12 @@ export function makeAuthenticatedQueries (userId, fetchOne, fetchMany) {
     post: (root, { id }) => fetchOne('Post', id),
     posts: (root, args) => fetchMany('Post', args),
     people: (root, args) => fetchMany('Person', args),
+    projects: (root, { context, contextId, viewerId }) => {
```
lorenjohnson, line 168: In terms of the graphql modeling for this particular feature I'd prefer to see `projects` as a resolver on Person e.g. `person (id: $id) { projects { ... }  }`. There may be a case in the future for a root level 'projects' resolver, but I think for the Network and Community landing pages I would apply the same log (i.e. a `projects` resolver on Community and Network, etc). There is no hard and fast rule here, and graphql is a fairly open pattern, but the context for the projects query here is set by a query already going out for the person/member being viewed.

api/models/Post.js
lorenjohnson: Had same question on this one.

lorenjohnson (changes requested):
Looks good overall! Notes are: What Tibet said on the queries, and regardless of whether we keep this as a root level `projects` resolver, or push it into `person` I think it'd be a good thing for us all to cohere around as a team (without holding-up this work).

api/models/Post.js
wushenshen: I didn't realize you could do `this.query(.....).fetchAll()` - will update

api/graphql/index.js
wushenshen: ok sounds good, I moved it into `projectsForContext`

api/models/Post.js
wushenshen: keeping it as * since it's the main query now

api/graphql/index.js
tibetsprague: i think i agree with Loren, but until we decide, this is ok

tibetsprague: i guess it depends on whether it makes sense to load the projects as part of 1 big query or whether it might be more efficient to load that section separately with a separate query. i dont have a strong opinion. Sticking with current norms we would load through the person resolver, though i think that query might be getting perhaps too large... All in all I would lean towards Loren's suggestion

lorenjohnson:
Yes, but the current resolvers don't generate one big query. For instance
the current `person` query called for the Profile page generates a bunch of
independent queries to postgres including one for the user and
separate ones for `comments`, `posts`, etc. What is at stake I think is the
shape of the graphql schema / our API.

The graphql layer is where we get to play with how and what queries go out
and caching and such. Currently some of the opportunities for optimization
are patterned away in the grapqhl-bookshelf-bridge + makeResolvers stack,
but we can over time replace that implementation or otherwise make
interventions to optimize performance at that layer. Being clear about our
schema design now will maintain those opportunities.

In this case, as I think is generally true with graphql API design,
additional context for queries is a good thing (e.g. `person (personId) {
projects }` vs `projects (personId) { }`) in terms of leveraging the power
of the graphql pattern in said ways as well as for creating coherency
(assuming a entity based model vs view based modeling).

wushenshen:
@lorenjohnson @tibetsprague updated - projects is now a resolver on Person

lorenjohnson (approved):
👍 This looks really tidy and ready to go to me!

tibetsprague (approved):

PR #562, opened 2020-11-20
Skills to learn

tibetsprague:
Don't really love this solution, but after many attempts this was the easiest/best way i could come up with
Fixes https://github.com/Hylozoic/hylo-node/issues/556

### DIFF
--- a/api/graphql/mutations/index.js
+++ b/api/graphql/mutations/index.js
@@ -147,7 +147,7 @@ export function unlinkAccount (userId, provider) {
   .then(() => ({success: true}))
 }
 
-export async function addSkill (userId, name) {
+export async function addSkill (userId, name, type) {
   name = trim(name)
   if (isEmpty(name)) {
     throw new Error('Skill cannot be blank')
@@ -165,7 +165,7 @@ export async function addSkill (userId, name) {
   }
 
   try {
-    await skill.users().attach(userId)
+    await skill.users().attach({ user_id: userId, type })
   } catch (err) {
     if (!err.message || !err.message.includes('duplicate')) {
       throw err
--- a/api/graphql/mutations/index.js
+++ b/api/graphql/mutations/index.js
@@ -147,7 +147,7 @@ export function unlinkAccount (userId, provider) {
   .then(() => ({success: true}))
 }
 
-export async function addSkill (userId, name, type) {
+async function createSkill(name) {
   name = trim(name)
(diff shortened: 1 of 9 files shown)


api/graphql/mutations/index.js
```diff
@@ -147,7 +147,7 @@ export function unlinkAccount (userId, provider) {
   .then(() => ({success: true}))
 }
 
-export async function addSkill (userId, name) {
+export async function addSkill (userId, name, type) {
   name = trim(name)
```
lorenjohnson, line 151: Could check for valid type here...

lorenjohnson:
Hey @tibetsprague I've came back and looked at this now a couple times. I continue to be feel it'd be best to keep this flat leaving the Skill `type` handling entirely here on the backend, sticking with `person { skillsToLearn {} }`, and creating an `addSkillToLearn` mutation (vs `addSkill (type)`) in the related graphql schema and resolvers. 

Leaving the abstraction exposed could have its use, but I think as it passes through the layers it's causing more confusion than utility at this stage. 

Thoughts?

tibetsprague:
> Hey @tibetsprague I've came back and looked at this now a couple times. I continue to be feel it'd be best to keep this flat leaving Skill `type` handling entirely here on the backend, sticking with `person { skillsToLearn {} }`, and creating an `addSkillToLearn` mutation (vs `addSkill (type)`) in the related graphql schema and resolvers.
> 
> Leaving the abstraction exposed could have its use, but I think as it passes through the layers it's causing more confusion than utility at this stage.
> 
> Thoughts?

yeah that makes sense for now too, just simplifies a bit more

tibetsprague:
@lorenjohnson ready for new review

api/graphql/schema.graphql
```diff
@@ -869,6 +873,7 @@ type CommunityQuerySet {
 type Skill {
   id: ID
   name: String
+  type: Int
```
lorenjohnson, line 876: I would keep the `type` out as it's not being used. If a differentiation is needed in type then I'd carry through treating this as a separate type and make a `SkillToLearn` and `SkillsToLearnQuerySet`.

lorenjohnson (approved):
👍  Overall looks great! 

I do think for clarity it'd be best to leave Skill#type entirely out of the graphql side, leaving it there otherwise may confuse or misdirect a future developer--but that's a nitpick at this point.

PR #566, opened 2020-12-01
Add query for a person's upcoming events

### LINKED ISSUE #557: Query for upcoming events for a person

wushenshen:
Closes #557

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -112,6 +112,7 @@ export default async function makeModels (userId, isAdmin) {
         'memberships',
         'moderatedCommunityMemberships',
         'locationObject',
+        {eventsAttending: {querySet: true}},
         {posts: {querySet: true}},
         {projects: {querySet: true}},
         {comments: {querySet: true}},
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Add eventsAttending to the User model and Person type
 
 ## [1.4.3]
 ### Added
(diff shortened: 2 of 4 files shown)


api/graphql/makeModels.js
```diff
@@ -111,6 +111,7 @@ export default async function makeModels (userId, isAdmin) {
         'memberships',
         'moderatedCommunityMemberships',
         'locationObject',
+        {events: {querySet: true}},
```
tibetsprague, line 114: i think we should probably call this eventsAttending

lorenjohnson: I agree and would want this reflected in the model too, if that is in fact what is being queried. I'm not actually clear. Is the `User#events` relation only pulling events to which I'm attending or also events I've merely created? 

If I remember correctly I'm always added to the event group ("attending") when I create an event. If that is the case I think the query and this naming is correct.

lorenjohnson (approved):
👍  LGTM --- pending clarification on the what `User#events` queries. If it's returning all events regardless of whether I created it and/or am attending, then we're good. If there is a distinction here we're missing then I think it needs another look.

api/graphql/makeModels.js
wushenshen: @lorenjohnson That's correct, the event creator is also added to the event group, and therefore "attending"

wushenshen:
> 👍 LGTM --- pending clarification on the what `User#events` queries. If it's returning all events regardless of whether I created it and/or am attending, then we're good. If there is a distinction here we're missing then I think it needs another look.

No distinction - all events regardless of whether the user created it and/or is attending.

api/graphql/makeModels.js
tibetsprague: @wushenshen so lets call it eventsAttending then, right? I think if a creator of an event says they are no longer attending the event it should disappear from the list on their profile

wushenshen:
@tibetsprague updated the naming and logic to pull only upcoming events the person is attending. fyi @lorenjohnson

Issue #559, opened 2020-11-18
Daily digest emails often not getting sent


tibetsprague:
complainers:
hkharlan@gmail.com
michaelfmargolies@gmail.com

Check this error: https://rollbar.com/hylo_dev/hylo-node/items/2882/

tibetsprague:
Specific community mike has not been getting digests from is https://www.hylo.com/c/d3

tibetsprague:
so sendwithus does confirm that Mike Margolies has not received a digest email in the last 30 days

PR #568, opened 2020-12-04
Fix bug in sending daily digests

tibetsprague:
Hopefully fixes https://github.com/Hylozoic/hylo-node/issues/559

I suspect this is new from the backend updates though and might not solve Mike's lack of digests?

### DIFF
--- a/cron.js
+++ b/cron.js
@@ -11,14 +11,14 @@ const savedSearches = require('./lib/community/digest2/savedSearches')
 
 const sendAndLogDigests = type =>
   digest2.sendAllDigests(type)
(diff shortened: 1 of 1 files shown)


wushenshen (approved):
Based on [the documentation](http://bluebirdjs.com/docs/api/tap.html) I believe this accomplishes the same thing, fine with trying it though

tibetsprague:
Yeah .tap is not a function anymore so it’s breaking. This replaces it 

Sent by m

lorenjohnson (commented):
👍LGTM

PR #574, opened 2020-12-11
548 custom org affiliations

### LINKED ISSUE #548: User profile addition - custom organizational affiliations backend
Should be able to add more than one with a role name, org name, and URL. 

New table user_affiliations with columns: user_id (ref), role (string), org_name (string) and url (string)

wushenshen:
Closes #548 

Related: https://github.com/Hylozoic/hylo-evo/pull/723


tibetsprague (approved):
LGTM!

PR #578, opened 2020-12-21
Easier initialization

jasonprado:
I tried to spin up hylo-node following the documentation but found it a little cumbersome. Some projects I've worked with have started using Docker to standardize their development environment and to make it very quick and simple to get started on the project.

I created a Dockerfile and docker-compose.yml file that build an environment for running hylo-node and also bring up Postgres/Redis servers locally. These files can be used on their own or with VSCode.

Using the [VSCode Remote Containers feature](https://code.visualstudio.com/docs/remote/containers-tutorial), VSCode will automatically create a development environment when opening a repository with a `.devcontainer/` directory. This means that a new developer basically just has to check out the repo and open it to be able to run a local version of the server.


jasonprado:
Let me know if this is a direction you find useful, or if you'd like to discuss more. Thanks!

pospi:
Really appreciate this addition, as someone new to the project who doesn't like polluting their host OS :D

I have gotten this running locally via a new config file that only boots the backing services (see above). This gives me a more lightweight development environment for my somewhat "inbetween" situation where I have node installed in the host but would prefer other services to be installed in a container. Added some instructions to your new readme about this as well.

I think this is complete enough to merge as-is, but ideally we could get something better in place for initialising the database. I think it should be possible to place the shell commands you have under "initializing the database" into a script and mount them into the postgres container via a volume (or just add a Dockerfile and copy them inside). I will need to do something like that for my setup since I don't have the Postgres CLI tools installed in my host machine, where I would need to run such commands from.

Would be very grateful if you'd like to take that on @jasonprado but if not it's likely I will add that in the coming days.

tibetsprague (approved):
Thanks @jasonprado ! Love this.

PR #579, opened 2020-12-25
Docker env seeding

pospi:
Update to #578 that allows DB initialisation commands to be executed by developers opting not to install Postgres tooling locally. May also avoid some potential issues with the existing setup script- unsure if `db` and `localhost` both resolve to the correct server in the VSCode dev env; but this addition makes such a concern explicit.


tibetsprague (approved):
LGTM

PR #581, opened 2021-01-11
update url validation when creating a user affiliation

wushenshen:
Small update to `createAffiliation` to only validate populated urls

### DIFF
--- a/api/graphql/mutations/affiliation.js
+++ b/api/graphql/mutations/affiliation.js
@@ -5,15 +5,16 @@ export async function canDeleteAffiliation (userId, affiliationId) {
   return affiliation.get('user_id') === userId
 }
 
-export async function createAffiliation (userId, { role, preposition, orgName, url }) {
-  if (!validator.isURL(url)) throw new Error(`Please enter a valid URL.`)
+export async function createAffiliation (userId, { role, preposition, orgName, url = '' }) {
+  const formattedUrl = 'https://' + url
+  if (url.length > 0 && !validator.isURL(formattedUrl)) throw new Error(`Please enter a valid URL.`)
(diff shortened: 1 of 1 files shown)


api/graphql/mutations/affiliation.js
```diff
@@ -5,15 +5,16 @@ export async function canDeleteAffiliation (userId, affiliationId) {
   return affiliation.get('user_id') === userId
 }
 
-export async function createAffiliation (userId, { role, preposition, orgName, url }) {
-  if (!validator.isURL(url)) throw new Error(`Please enter a valid URL.`)
+export async function createAffiliation (userId, { role, preposition, orgName, url = '' }) {
+  const formattedUrl = 'https://' + url
```
tibetsprague, line 9: what if the person adds http:// or https:// themselves? copying and pasting a URL will do this for example. definitely need to check for that and not add it a second time

wushenshen: I was thinking that in that case it would throw the "Please enter a valid URL." error and the user would correct it. The input is prepended with "https://" and they can't delete past that

tibetsprague: ah gotcha, hmm. ok I could imagine this being tricky for some people, but i think its good enough for now. ideally the front-end would auto format it based on what they paste in and the backend would validate

tibetsprague: actually, no matter what I think its important that the backend handle the validation well and so checks for the http(s) first and only adds it if its missing

wushenshen: okay, I added a new function for that

tibetsprague (approved):

Issue #570, opened 2020-12-09
create DB tables for Landing Page widgets


cpolitano:
- [x] widgets table with all the widget types available (columns: name [string])
- [x]  pivot table group_widgets that includes group_id, widget_id, settings [jsonb] (such as text in text block), visibility [bool], and order on page [integer]
- [x] GraphQL resolver for querying widgets by group
- [x] GraphQL mutation for updating widget fields for a group
- [ ] When a new group is created, create records in the pivot table for all the widgets in the default order below

Types of widgets: ([illustrated here](https://www.figma.com/file/qy20FQN1q42OOPnGNSbfVh/Community-Home?node-id=0%3A1))
text_block
announcements
active_members
requests_offers
posts
community_topics
events
project_activity
group_affiliations
map

PR #586, opened 2021-01-17
Create widgets and group widgets tables

wushenshen:
#570 

Submitting this now for some preliminary feedback - the mutation resolver for updating widgets still needs to be done, and the full feature of adding widgets when a group is created


api/models/Community.js
```diff
@@ -86,6 +86,10 @@ module.exports = bookshelf.Model.extend(merge({
     })
   },
 
+  widgets: async function () {
+    return await Widget.fetchForCommunity(this.id)
```
tibetsprague, line 90: why not use a normal hasMany through relationship here?

api/models/GroupWidget.js
```diff
@@ -0,0 +1,48 @@
+import HasSettings from './mixins/HasSettings'
+
+module.exports = bookshelf.Model.extend(Object.assign({
+  tableName: 'group_widgets',
+
+  community: async function () {
+    return await Community.find(this.get('group_id'))
```
tibetsprague, line 7: nice, thanks for anticipating the move to groups

tibetsprague: this could be a belongsTo right?

api/models/Widget.js
```diff
@@ -0,0 +1,18 @@
+module.exports = bookshelf.Model.extend({
+  tableName: 'widgets',
+  requireFetch: false,
+}, {
+  fetchForCommunity: async function(community_id) {
+    const query = `select gw.id, w.name, gw.settings, gw.is_visible, gw.order from group_widgets gw
+    left join groups as g on g.id = gw.group_id
+    left join widgets as w on w.id = gw.widget_id
+    where g.group_data_id = ${community_id} and g.group_data_type = 1`
+    const result = await bookshelf.knex.raw(query)
```
tibetsprague, line 10: why did you want/need a raw query for this instead of using bookshelf?

PR #595, opened 2021-01-27
580 threaded comments

pospi:
Various fixes, field name changes and updates to support multi-level threaded comments in the backend.


pospi:
1ebc580 is noteworthy, in light of 472fd11 (which I found a bit scary but we all were happy with). Given that we haven't got a working test suite on the backend yet, I think such potentially far-reaching set of changes warrants a really exhaustive set of user testing. Probably best deployed along with the platform, redux-orm etc changes so that we can do all the "whole app" testing at once?

api/models/Post.js
```diff
@@ -67,7 +67,10 @@ module.exports = bookshelf.Model.extend(Object.assign({
   },
 
   comments: function () {
-    return this.hasMany(Comment, 'post_id')
+    return this.hasMany(Comment, 'post_id').query({ where: {
+      'comments.active': true,
+      'comments.comment_id': null
```
tibetsprague, line 72: this seems fine for now but abstractly it makes sense that the comments() relationship should return all comments on the post to me, and maybe a parameter would say to only return top level comments? dont need to change this now just wanted to flag that

tibetsprague (approved):
Looks good! Seems like it would be fine to merge this into dev now right? what did you think it might mess up again? seems fine to just return top level comments only because thats all we have right now. On the other hand this probably wont go out until holonic architecture because of the need to implement on mobile... so we could also just wait. thoughts?

tibetsprague:
already merged into holonic architecture

PR #597, opened 2021-02-01
Fix issue where people are seen incorrectly as a moderator

tibetsprague:
Also add person resolver to memberships to fix some bugs on the front-end

Fixes https://github.com/Hylozoic/hylo-node/issues/596


lorenjohnson (approved):
👍  Well that feels good. A little change to make things much better. Love it.

PR #624, opened 2021-03-26
590 correct posts for context

tibetsprague:
Ensure we show the right set of posts and groups for each context (public, all and group)

### DIFF
--- a/api/graphql/filters.js
+++ b/api/graphql/filters.js
@@ -138,10 +138,9 @@ export const postFilter = (userId, isAdmin) => relation => {
       q.where(tableName + '.is_public', true)
     } else if (!isAdmin) {
       // Only show posts that are public or posted to a group the user is a member of
-      q.join('groups_posts as gp', 'gp.post_id', 'posts.id')
       q.where(q3 => {
         const selectIdsForMember = Group.selectIdsForMember(userId)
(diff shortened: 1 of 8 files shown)


lorenjohnson (commented):
I've spent some time with it and I am struggling to fully grok what was accomplished when considering the related front-end changes. 

Let's take 20 mins and walk through this together at co-work today. My questions as of now are:  

1) Where does the new `viewPosts` ever get called? I don't see it in (what I thought was) the related PR for evo: https://github.com/Hylozoic/hylo-evo/pull/851/files

2) Why `group { viewPosts {} }` vs changing the behavior of `group { posts {} }`? I don't see where we're currently using `group { posts `} in the frontend independent of the "Stream" usage. 

3) The distinction between `makePublic` and `makeAuthenticatedQueries` (predating your changes in this PR) comment that they are distinct by a logged-in vs non-logged-in user yet I don't think we have any views for a non-authenticated user. The dividing line here is just between groups/posts which I can see that I'm a member of vs those I can't. Right?

---

Regardless, I think I'm just not fully getting the picture so take the questions with a grain of salt, but they can be where we pick-up from if we can look at this together for a moment.

tibetsprague:
Yeah lets talk through today

PR #627, opened 2021-04-02
603 group data export

pospi:
Mostly implemented for MVP of https://github.com/Hylozoic/hylo-evo/issues/796, wanted to submit this for review and also because the last query there is complicated and I'd like some help with it-

- [x] Export groups the person is a part of (that the moderator can see)

Otherwise, the only other thing to really check is the formatting. I've gone with comma-delimited cell separators which I *think* is standard- the output library quotes fields safely automatically.


tibetsprague:
yeah comma separated is what we want

api/controllers/ExportController.js
```diff
@@ -0,0 +1,136 @@
...
+ */
+async function exportMembers(groupId, req, res) {
+  const users = await new Group({ id: groupId })
+                  .members()
+                  .fetch()
+
+  const results = []
+
+  // iterate over all group members
+  await Promise.all(users.map((u, idx) => {
```
tibetsprague, line 48: this is where i would use Promise.map

api/controllers/ExportController.js
```diff
@@ -0,0 +1,136 @@
...
+    // pluck core user data into results
+    results.push(u.pick([
+      'name', 'contact_email', 'contact_phone', 'location', 'avatar_url', 'tagline', 'bio',
```
tibetsprague, line 51: lets also pull in the individual location fields from the locations table

api/controllers/ExportController.js
```diff
@@ -0,0 +1,136 @@
...
+
+      // skills to learn
+      u.skillsToLearn().fetch()
+        .then(skills => {
+          results[idx].skills_to_learn = accumulatePivotCell(skills, renderSkill)
+        }),
+
+      // Join request questions & answers
+      u.joinRequests().where({ group_id: groupId, status: JoinRequest.STATUS.Accepted }).fetch()
+        .then(jrs => Promise.all(jrs.map(jr =>
```
tibetsprague, line 79: hmm, shouldnt there only ever be one accepted join request? lets just sort by created_at DESC and choose the one most recent accepted one (its possible they could join then leave the group then request to join again and be accepted again)

api/controllers/ExportController.js
```diff
@@ -0,0 +1,136 @@
...
+  await Promise.all(users.map((u, idx) => {
+    // pluck core user data into results
+    results.push(u.pick([
+      'name', 'contact_email', 'contact_phone', 'location', 'avatar_url', 'tagline', 'bio',
+      'url', 'twitter_url', 'facebook_url', 'linkedin_url'
+    ]))
+
+    // return combined promise to load all dependent user data and
+    // assign final child query results back onto matching result objects upon completion
+    return Promise.all([
```
tibetsprague, line 57: gotta say i think this whole thing would look way cleaner if you just use awaits everywhere. you can still use Promise.map to walk through one by one and use async/await in the callback of the map. but im ok with rolling with this if its too hard

api/controllers/ExportController.js
```diff
@@ -0,0 +1,136 @@
...
+  }))
+
+  // send data as CSV response
+  output(res, results, [
+    'name', 'contact_email', 'contact_phone', 'location', 'avatar_url', 'tagline', 'bio',
+    { key: 'url', header: 'personal_url' },
+    'twitter_url', 'facebook_url', 'linkedin_url',
+    'skills', 'skills_to_learn',
+    'join_request_questions',
+    // :TODO: groups the person is a part of (that the moderator can see)
```
tibetsprague, line 102: the easiest way to do this is to load the user.memberships() and then apply the groupFilter from filters.js which filters out groups that the current user cant see. i think the way to do this is call `groupFilter(userId)(user.memberships()).fetch()` ? that just might do it

tibetsprague:
ok we have one more request if you are down @pospi. we want a second export option/button that exports the member data for the current group + all the members of child groups of the current group that the current user is also a moderator of. You can use Group.allChildGroups for that along with the moderator check

api/controllers/ExportController.js
pospi: If by this you meant the latitude and longitude, that's done. Don't think there was any other data in there that was relevant?

tibetsprague (approved):

PR #657, opened 2021-08-17
Hylo 611 update member directory export

thomasgwatson:
Successfully got member export lists to show up in my email

Separating out the changes to the request-group-questions-answers into another PR, will want to pair with you a little to get me rolling on that.

### DIFF
--- a/api/controllers/ExportController.js
+++ b/api/controllers/ExportController.js
@@ -6,6 +6,9 @@ module.exports = {
   groupData: async function (req, res) {
     const p = req.allParams()
 
+    const user = await new User({id: req.session.userId} )
+                  .fetch({ columns: ['email']})
+
     if (!p.groupId) {
       throw new Error("Please specify group ID")
     }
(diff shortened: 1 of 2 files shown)


api/controllers/ExportController.js
```diff
@@ -113,27 +117,41 @@ async function exportMembers(groupId, req, res) {
...
+    const partialGroupId = groupId.slice(0,7)
+    const formattedDate = new Date().toISOString().slice(0,10)
+    const buff = Buffer.from(output)
+    const base64output = buff.toString('base64')
+
+    Queue.classMethod('Email', 'sendExportMembersList', {
+      email:  email,
+      files: [
+        {
+          id: `members-export-${partialGroupId}-${formattedDate}.csv`,
```
tibetsprague, line 148: we could use group name here instead of id

tibetsprague (approved):

api/controllers/ExportController.js
thomasgwatson: Will address in next PR

Issue #517, opened 2020-07-27
Get hylo-node tests passing


tibetsprague:
Make sure test run actually errors when tests fail.

Use Jest?

PR #662, opened 2021-08-31
All tests working! Many bugs fixed!

tibetsprague:
All Hylo Node tests now pass. Fixes https://github.com/Hylozoic/hylo-node/issues/517

Also a number of bugs were fixed:
- Various issues with unread counts for chat threads and new post counts for group memberships. Fixes https://github.com/Hylozoic/hylo-evo/issues/937
- Duplicate threads getting created when messaging the same group of people
- Correctly export twitter name


api/controllers/CommentController.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
     )
   },
   createBatchFromEmailForm: function (req, res) {
+    // TODO: fix
```
thomasgwatson, line 41: Fix for now or later?

api/models/User.js
```diff
@@ -97,12 +97,11 @@ module.exports = bookshelf.Model.extend(merge({
...
 
   posts: function () {
-    return this.belongsToMany(Post).through(PostUser).query(q => q.where(function () {
-      this.where('type', null).orWhere('type', '!=', Post.Type.THREAD)
-    }))
+    return this.hasMany(Post).query(q => q.where('type', '!=', Post.Type.THREAD))
   },
 
   projects: function () {
+    // TODO: fix
```
thomasgwatson, line 104: Now or later?

thomasgwatson (approved):
Checked it over as best I could. Just checking in on a few of the TODOs as to whether that was TODO now or later

api/controllers/CommentController.js
tibetsprague: oh yeah this is an eventually someday if we ever bring back this feature... not really important enough to track with a ticket yet just wanted to flag in the code that this functionality is not totally working

api/models/User.js
tibetsprague: later, whenever we get back to the projects functionality in a real way

PR #683, opened 2021-09-24
Update to node 14.7.6, babel 7 and node-redis 4

tibetsprague:
Plus some other minor package updates


api/services/RedisClient.js
```diff
@@ -1,9 +1,10 @@
-import redis from 'redis'
-Promise.promisifyAll(redis.RedisClient.prototype)
-Promise.promisifyAll(redis.Multi.prototype)
```
thomasgwatson, line 3: Are we being forced to update promises/.then to async/await or is that by choice?

thomasgwatson (commented):
lgtm

api/services/RedisClient.js
tibetsprague: by choice but also because I wanted features from the latest version of reds which switches from callbacks to async/await, which is sooo much better

lorenjohnson (commented):
👍  LGTM! This is such a welcome set of changes. Disclaimer: I've done a code review directly only, I've not ran this yet as my backend and tested things that way.

PR #691, opened 2021-10-21
Update valid Socket allowedOrigins to include Mobile

lorenjohnson:


lorenjohnson:
Forgive me: I've updated the CHANGELOG and version and am going to go ahead and merge / deploy. This should make Mobile sockets good now and I've tested the change on Staging as well as it's a trivial change.

PR #696, opened 2021-10-29
Add second type of duplicate location record check

thomasgwatson:


api/graphql/mutations/location.js
```diff
@@ -1,20 +1,33 @@
...
+        address_street: data.address_street,
+        city: data.city || null,
+        locality: data.locality || null,
+        neighborhood: data.neighborhood || null,
+        postcode: data.postcode || null,
+        country: data.country || null
+      }
+    }
+  }
+  if (data.full_text) {
```
tibetsprague, line 20: this needs to be an else if

tibetsprague (approved):

Issue #620, opened 2021-03-24
Mutations for a user to deactivate & delete themselves


tibetsprague:
Deactivation:
- Back-end: set their active column to false. Log them out and delete their session. 
- If they try to log back in say this will re-activate your account, first asking them if they are sure they want to do that. 
- What should we show for posts and messages with deactivated users? Leave content for now we think... make sure people know it will be left up
- Show them in membership directory? No

Delete account:
- We don't delete the posts, we just delete the content of the post and change it to say Deleted Post, but leave comments on the post by other people. And change comments to say Deleted Comment. We will likely need a user that is called Deleted User that we point deleted posts and comments too as the creator/owner.
- What about upvotes? Could change to point to the Deleted User row in the users table, so we still keep the count the same...?
- Do we delete the groups they created? If other moderators we leave up as normal? If they were the only moderator in the group we pause/deactivate the group. Don't delete it but no more activity can happen in it. (This will also cause it to disappear for members of the group.) Do we need to send them a notification saying the group was deactivated?

thomasgwatson:
"What about upvotes? Could change to point to the Deleted User row in the users table, so we still keep the count the same...?"

Normally I assume a user can't have more than one upvote on a post; so I'm assuming that we can only do this trick once before basically start removing the upvote count. Perhaps it is more sensible to just delete the upvote (or emoji react in the future)

Perhaps more importantly for future thinking, what do we do with a deleted users `votes` in important proposals? What are the implications of having a vote record where the numbers of votes change over time with deleted accounts? Interesting fodder for later discussions.

thomasgwatson:
Users before deletion

![image](https://user-images.githubusercontent.com/5193218/145127700-8eefc8f2-7945-407c-bc38-d993c396599b.png)

Deleted one user, got this error:
![image](https://user-images.githubusercontent.com/5193218/145128248-6e7b54e3-27f8-44cb-ba01-02f29ca6f1d4.png)

thomasgwatson:
After updating the `deletedUser` record with a `name` we get

![image](https://user-images.githubusercontent.com/5193218/145129273-89eac52c-57c1-4360-bffc-ccf9614b60fc.png)

tibetsprague:
just realizes there's another thing we probably need to do: delete all the photos on AWS from the user's avatar and posts

thomasgwatson:
[...]

PR #705, opened 2021-11-16
Node 620 deactivate user

thomasgwatson:
What else does this need to work? This would be my first mutation, so curious to know if I have missed anything.

For #620

### DIFF
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -27,7 +27,6 @@ import {
   createSavedSearch,
   createTopic,
   deactivateUser,
-  deleteUser,
   declineJoinRequest,
   deleteAffiliation,
   deleteComment,
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -26,7 +26,8 @@ import {
   createProjectRole,
   createSavedSearch,
   createTopic,
-  deactiveUser,
+  deactivateUser,
+  deleteUser,
   declineJoinRequest,
   deleteAffiliation,
   deleteComment,
@@ -57,7 +58,7 @@ import {
   registerStripeAccount,
   reinviteAll,
   rejectGroupRelationshipInvite,
-  reactiveUser,
+  reactivateUser,
   removeMember,
   removeModerator,
   removePost,
@@ -246,7 +247,7 @@ export function makeMutations (sessionId, userId, isAdmin) {
 
     createTopic: (root, { topicName, groupId, isDefault, isSubscribing }) => createTopic(userId, topicName, groupId, isDefault, isSubscribing),
 
-    deactivateMe: (root, { sessionId, userId }) => deactiveUser({ sessionId, userId }),
+    deactivateMe: (root, { sessionId, userId }) => deactivateUser({ sessionId, userId }),
 
     declineJoinRequest: (root, { joinRequestId }) => declineJoinRequest(userId, joinRequestId),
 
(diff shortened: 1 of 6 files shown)


api/graphql/index.js
```diff
@@ -65,7 +67,6 @@ import {
   resendInvitation,
   respondToEvent,
   subscribe,
-  toggleGroupWidgetVisibility,
```
thomasgwatson, line 68: linter picked this up as unused, so I removed.

api/models/User.js
```diff
@@ -161,6 +161,15 @@ module.exports = bookshelf.Model.extend(merge({
     .digest('hex')
   },
 
+  reactivate: function () {
+    return this.set({ active: true })
+  },
+
+  deactivate: async function (sessionId) {
+    await User.clearSessionsFor({sessionId, userId: this.id})
```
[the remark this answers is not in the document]
tibetsprague: also I still think we want to delete the current session too:     req.session.destroy(), maybe that happens in the mutation?

api/models/User.js
```diff
@@ -161,6 +161,15 @@ module.exports = bookshelf.Model.extend(merge({
     .digest('hex')
   },
 
+  reactivate: function () {
+    return this.set({ active: true })
```
tibetsprague, line 165: I thought that .set didn't actually do a .save()? We usually do .save({ active: true })

tibetsprague (changes requested):

api/models/User.js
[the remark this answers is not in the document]
thomasgwatson: How do I play with the request in the graphQL context?

I only see references to `req.session` in https://github.com/Hylozoic/hylo-node/blob/d073057fe5f1a1053c1c61b517eaf1b5d634aa85/api/graphql/index.js#L371 and don't really see other mutations or queries dealing directly with the `req`. Any idea on what the convention is for this?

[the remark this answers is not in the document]
tibetsprague: I dont actually... what if you actually just call logout on the front-end after the deactivate call?

api/graphql/schema.graphql
```diff
@@ -3,6 +3,7 @@ scalar Date
 # The currently logged-in person.
 type Me {
   id: ID
+  active: Boolean
```
tibetsprague, line 6: is there ever a time we would return a deactivated me?

tibetsprague (approved):

api/graphql/index.js
```diff
@@ -245,6 +246,8 @@ export function makeMutations (sessionId, userId, isAdmin) {
 
     createTopic: (root, { topicName, groupId, isDefault, isSubscribing }) => createTopic(userId, topicName, groupId, isDefault, isSubscribing),
 
+    deactivateMe: (root, { sessionId, userId }) => deactivateUser({ sessionId, userId }),
```
tibetsprague, line 249: shouldnt need to pass in userId because for deactivateMe we always want to deactivate the current user

api/graphql/index.js
```diff
@@ -298,6 +301,8 @@ export function makeMutations (sessionId, userId, isAdmin) {
     processStripeToken: (root, { postId, token, amount }) =>
       processStripeToken(userId, postId, token, amount),
 
+    reactivateMe: (root, { userId }) => reactivateUser({ userId }),
```
tibetsprague, line 304: shouldnt need to pass in userId, always current user

tibetsprague (changes requested):

PR #706, opened 2021-11-22
Node 620 delete user (includes prior deactivate user commits)

### LINKED ISSUE #620: Mutations for a user to deactivate & delete themselves

thomasgwatson:
Closes #620

### DIFF
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -27,7 +27,6 @@ import {
   createSavedSearch,
   createTopic,
   deactivateUser,
-  deleteUser,
   declineJoinRequest,
   deleteAffiliation,
   deleteComment,
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -26,7 +26,8 @@ import {
   createProjectRole,
   createSavedSearch,
   createTopic,
-  deactiveUser,
+  deactivateUser,
+  deleteUser,
   declineJoinRequest,
   deleteAffiliation,
   deleteComment,
@@ -57,7 +58,7 @@ import {
   registerStripeAccount,
   reinviteAll,
   rejectGroupRelationshipInvite,
-  reactiveUser,
+  reactivateUser,
   removeMember,
   removeModerator,
   removePost,
@@ -246,7 +247,7 @@ export function makeMutations (sessionId, userId, isAdmin) {
 
     createTopic: (root, { topicName, groupId, isDefault, isSubscribing }) => createTopic(userId, topicName, groupId, isDefault, isSubscribing),
 
-    deactivateMe: (root, { sessionId, userId }) => deactiveUser({ sessionId, userId }),
+    deactivateMe: (root, { sessionId, userId }) => deactivateUser({ sessionId, userId }),
 
     declineJoinRequest: (root, { joinRequestId }) => declineJoinRequest(userId, joinRequestId),
 
@@ -301,7 +302,7 @@ export function makeMutations (sessionId, userId, isAdmin) {
     processStripeToken: (root, { postId, token, amount }) =>
       processStripeToken(userId, postId, token, amount),
 
-    reactivateMe: (root, { userId }) => reactiveUser({ userId }),
+    reactivateMe: (root, { userId }) => reactivateUser({ userId }),
 
     regenerateAccessCode: (root, { groupId }) =>
       regenerateAccessCode(userId, groupId),
--- a/api/graphql/index.js
+++ b/api/graphql/index.js
@@ -27,6 +27,7 @@ import {
   createSavedSearch,
   createTopic,
   deactivateUser,
+  deleteUser,
   declineJoinRequest,
   deleteAffiliation,
   deleteComment,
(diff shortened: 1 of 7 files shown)


api/graphql/index.js
```diff
@@ -239,12 +241,14 @@ export function makeMutations (sessionId, userId, isAdmin) {
...
     createSavedSearch: (root, { data }) => createSavedSearch(data),
 
-    joinGroup: (root, {groupId}) => joinGroup(groupId, userId),
+    joinGroup: (root, { groupId }) => joinGroup(groupId, userId),
 
     joinProject: (root, { id }) => joinProject(id, userId),
 
     createTopic: (root, { topicName, groupId, isDefault, isSubscribing }) => createTopic(userId, topicName, groupId, isDefault, isSubscribing),
 
+    deactivateMe: (root, { id }) => deactivateUser({ sessionId, userId: id }),
```
tibetsprague, line 250: i just noticed that you are passing in id here, shouldnt you just always use userId? we dont want to let people deactivate other users, at least through this deactivateMe call

api/graphql/index.js
```diff
@@ -257,6 +261,8 @@ export function makeMutations (sessionId, userId, isAdmin) {
 
     deleteGroupTopic: (root, { id }) => deleteGroupTopic(userId, id),
 
+    deleteMe: (root, {id}) => deleteUser({ sessionId, userId: id }),
```
tibetsprague, line 264: i just noticed that you are passing in id here, shouldnt you just always use userId? we dont want to let people delete other users, at least through this deleteMe call

api/graphql/index.js
```diff
@@ -298,6 +304,8 @@ export function makeMutations (sessionId, userId, isAdmin) {
     processStripeToken: (root, { postId, token, amount }) =>
       processStripeToken(userId, postId, token, amount),
 
+    reactivateMe: (root, { id }) => reactivateUser({ userId: id }),
```
tibetsprague, line 307: same

api/graphql/mutations/user.js
```diff
@@ -2,14 +2,34 @@ import request from 'request'
...
+  return { success: true }
+}
+
+export async function reactivateUser ({ userId }) {
+  const user = await User.find(userId, {}, false)
+  await user.reactivate()
+  return { success: true }
+}
+
+export async function deleteUser ({ userId, sessionId }) {
```
tibetsprague, line 27: the permission check should be that the mutation only allows for deleting the current user, but this function can take a userId for potential other use in the future

api/graphql/schema.graphql
```diff
@@ -901,6 +902,8 @@ type Mutation {
   createProjectRole(projectId: ID, roleName: String): GenericResult
   createSavedSearch(data: SavedSearchInput): SavedSearch
   createTopic(topicName: String, groupId: ID, isDefault: Boolean, isSubscribing: Boolean): Topic
+  deactivateMe(id: ID): GenericResult
```
tibetsprague, line 905: no id needed

api/models/User.js
```diff
@@ -161,6 +161,72 @@ module.exports = bookshelf.Model.extend(merge({
...
+
+      - zero out content of their posts and comments
+      - change the ownership of their posts and comments to the deletedUser record
+      - remove other references to their user_id
+      - delete their user record!
+    */
+
+    const query = `
+    BEGIN;
+    UPDATE posts SET name = 'Post by deleted user', description = '', active = false, location = NULL, location_id = NULL WHERE user_id = ${this.id};
```
tibetsprague, line 188: will active = false make it invisible to the front-end? is that what we want? we may want people to still see the deleted post with the existing comment thread on it

api/models/User.js
```diff
@@ -161,6 +161,72 @@ module.exports = bookshelf.Model.extend(merge({
...
+    DELETE FROM thanks WHERE thanked_by_id = ${this.id};
+    DELETE FROM notifications WHERE activity_id in (select id from activities WHERE reader_id = ${this.id});
+    DELETE FROM notifications WHERE activity_id in (select id from activities WHERE actor_id = ${this.id});
+    DELETE FROM activities WHERE actor_id = ${this.id};
+    DELETE FROM activities WHERE reader_id = ${this.id};
+
+    DELETE FROM contributions WHERE user_id = ${this.id};
+    DELETE FROM devices WHERE user_id = ${this.id};
+    DELETE FROM group_invites WHERE used_by_id = ${this.id};
+    DELETE FROM group_invites WHERE invited_by_id = ${this.id};
```
tibetsprague, line 208: this one im not sure about, should the invite be rescinded?

migrations/20211118103939_add_deleted_user_record.js
```diff
@@ -0,0 +1,14 @@
+
+exports.up = function (knex) {
+  return knex.raw(`
+    INSERT INTO users(email, first_name, last_name, active, email_validated, bio)
+    VALUES ('deleted@hylo.com', 'Deleted', 'User', false, true, 'This is the generic "deleted user" account');
```
tibetsprague, line 5: should we create this email account for some reason maybe?

tibetsprague (changes requested):

migrations/20211118103939_add_deleted_user_record.js
thomasgwatson: I was leaning towards no, but is there a compelling reason to do so? Will some functionality break if we don't have a real email associated with this account?

api/models/User.js
thomasgwatson: I think this is what I copied from the delete user DB procedure, so I don't have a strong opinion. If we don't want to rescind, I suppose we'll have to switch the `id` over to the deleted user.

api/models/User.js
thomasgwatson: Right, looks like it will hide them. I'll change.

api/graphql/index.js
thomasgwatson: Ahh right. I noticed that userId was the account I was logged into, and I didn't want to delete that account, so changed this. Will change it back

api/models/User.js
tibetsprague: hmm, maybe we should ask what people think, im not sure what is right here

migrations/20211118103939_add_deleted_user_record.js
tibetsprague: i cant think of anything that would break right now. i wonder if/how we can turn off notifications for this user? also do we need to add this user to the groups that posts are in? we will need to test the crap out of this all

tibetsprague (approved):

Issue #687, opened 2021-10-06
Use more secure JWTs for password reset and email verification


tibetsprague:
Add auth by JWT for password reset and email verification:
- tokens expire after a few hours

Issue #666, opened 2021-09-08
Verify people's email addresses


tibetsprague:
We agreed that you have to verify before you get in, this is what Slack and FB do now. 

You will get an email with a code to enter, and also a link option to click and confirm that way

PR #707, opened 2021-11-24
Add backend and APIs for verifying an email address

tibetsprague:
Add user_verification_codes table
Add JWT based auth, and using that for password resetting and email verify links

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1,8 +1,10 @@
+import jwt from 'jsonwebtoken'
+
 module.exports = {
   create: function (req, res) {
-    const { name, email, password } = req.allParams()
+    const { name, email, email_validated, password } = req.allParams()
 
-    return User.create({name, email: email ? email.toLowerCase() : null, account: {type: 'password', password}})
+    return User.create({name, email: email ? email.toLowerCase() : null, email_validated, account: {type: 'password', password}})
     .then(async (user) => {
       await Analytics.trackSignup(user.id, req)
       await req.param('login') && UserSession.login(req, user, 'password')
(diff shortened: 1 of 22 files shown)


tibetsprague:
Fixes https://github.com/Hylozoic/hylo-node/issues/666

tibetsprague:
Also fixes https://github.com/Hylozoic/hylo-node/issues/687

api/controllers/UserController.js
```diff
@@ -1,11 +1,15 @@
+import jwt from 'jsonwebtoken'
+
 module.exports = {
   create: function (req, res) {
-    const { name, email, password } = req.allParams()
+    const { name, email, email_validated, password } = req.allParams()
```
lorenjohnson, line 5: It smells a little funny to me to have `email_validated` provided from the frontend. It would seem better to only. check if the code is validated on the code entry screen, then send the code again from the frontend when the user is actually created and remove the email validation record at the point of user creation. Obviously this would require a small set of amendments on both this back end and the front end implementations, but overall I think that would be more secure and potentially slightly simpler overall. I am anxious as I'm sure you are to get this off our plate, but if you're agree and are willing to be make the changes I can amend mobile quickly.

lorenjohnson (commented):
This looks good overall, but consider my suggested amendment to the flow around the passing of `email_validated`

api/controllers/UserController.js
tibetsprague: yeah i could see that, i went this way because i actually thought maybe there is a future scenario where we want to allow people to create users without validated emails. so it was actually a choice to have the front-end handle the verification flow and tell us

tibetsprague: the one other advantage your version gives is that the code isnt deleted until the user is created so that the user can click the verify link multiple times and have it work (until it expires in 4 hours). This could be handy if a user clicks it on one device but then has to run off and then wants to verify on a different device shortly later. Not likely to happen a lot but maybe sometimes

tibetsprague: looking at it now, and feeling like this might be more of a refactor than i want to undertake right now. because i think it would mean changing how im handling the storing of emails (in the cookie)... honestly if its something you are excited to do id be happy to have you do it, otherwise id rather roll with what we have

lorenjohnson: All things considered, I'm with you. It seems like it could linger a long time but I've went ahead and added a ticket to the backlog to review this again in the mythical quite moment down the road 😌 https://github.com/Hylozoic/hylo-node/issues/717

lorenjohnson (approved):
👍🏼

PR #710, opened 2021-12-01
Automatically reactivate user if they log back in

### LINKED ISSUE #620: Mutations for a user to deactivate & delete themselves

thomasgwatson:
Addresses the issue of how to handle reactivation in the backend.

Closes #620

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -547,7 +547,8 @@ module.exports = bookshelf.Model.extend(merge({
     } else {
       q = User.where({ id })
     }
-    return q.where('users.active', active).fetch(options)
+    if (active) return q.where('users.active', active).fetch(options)
+    return q.fetch(options)
   },
 
   named: function (name) {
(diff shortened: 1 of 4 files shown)


api/models/User.js
```diff
@@ -547,7 +547,8 @@ module.exports = bookshelf.Model.extend(merge({
     } else {
       q = User.where({ id })
     }
-    return q.where('users.active', active).fetch(options)
+    if (active) return q.where('users.active', active).fetch(options)
```
tibetsprague, line 550: why do we need this? this means we couldnt pass in active: false to search for inactive users

thomasgwatson: I added the `active` parameter earlier in this work, so that the reactivate mutation would work. I made it only search `active = false` but it doesn't need to be that specific. The original behavior was searching amongst `active = true`, which I have preserved as the default behavior.

The changes I have added need to search both `active = true` and `active = false`, since it needs to return a user whether they are active or not

tibetsprague: its just weird to pass in active = false and have that mean to search for both active and non-active users, at least comment it

tibetsprague (approved):

PR #711, opened 2021-12-08
Switch to wiping user records, instead of deleting them

### LINKED ISSUE #620: Mutations for a user to deactivate & delete themselves

thomasgwatson:
Closes #620

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -171,19 +171,30 @@ module.exports = bookshelf.Model.extend(merge({
     return this.save({ active: false })
   },
 
+  deleteUserMedia: async function () {
+    const userId = this.get('id')
+    const userUrls = [this.get('banner_url'), this.get('avatar_url')]
+
+    const mediaUrls = await Media.findMediaUrlsForUser(userId)
+    const urls = mediaUrls.concat(userUrls)
+    Queue.classMethod('Media', 'deleteMediaByUrl', { urls })
+  },
+
   sanelyDeleteUser: async function ({ sessionId, transacting = {} }) {
-    Queue.classMethod('User', 'clearSessionsFor', { userId: this.get('user_id'), sessionId })
     /* 
       ### List of things to be done on account deletion ###
 
+      - Look up urls for all their possible uploads
+      - drop urls from external sources
+      - iterate through urls to delete each upload
+
       - zero out content of their posts and comments
-      - change the ownership of their posts and comments to the deletedUser record
       - remove other references to their user_id
-      - delete their user record!
+      - wipe their user record!
     */
 
-    // look up their uploads and delete
-    // Banner url, avatar url, user_external_data ???, and the media table?
+    await this.deleteUserMedia()
+    Queue.classMethod('User', 'clearSessionsFor', { userId: this.get('user_id'), sessionId })
 
     const query = `
     BEGIN;
@@ -554,7 +565,7 @@ module.exports = bookshelf.Model.extend(merge({
       })
     )
   },
-  
+
   find: function (id, options, activeFilter = true) {
     if (!id) return Promise.resolve(null)
     let q
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -207,10 +207,8 @@ module.exports = bookshelf.Model.extend(merge({
     DELETE FROM thanks WHERE thanked_by_id = ${this.id};
     DELETE FROM notifications WHERE activity_id in (select id from activities WHERE reader_id = ${this.id});
     DELETE FROM notifications WHERE activity_id in (select id from activities WHERE actor_id = ${this.id});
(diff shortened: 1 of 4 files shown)


api/models/User.js
```diff
@@ -183,30 +182,28 @@ module.exports = bookshelf.Model.extend(merge({
       - delete their user record!
     */
 
+    // look up their uploads and delete
```
thomasgwatson, line 185: Somewhat blocked on what to do here. I am assuming there is something happening outside of the node code that is putting files into a CDN, after they are inserted into s3. And our db records seem to mainly reference the CDN urls. 

So to delete, I have to reconstruct s3 keys from a file's CDN url? Is that possible? And then I have to delete the s3 object and also invalidate the CDN url record?

tibetsprague:
We just upload stuff to S3 and then cloud front sits in front of S3. The URLs for S3 and cloud front are the same, just with a different host. See getFinalUrl in lib/uploader/storage.js
So you can just find the same path in S3 and delete it.

api/models/User.js
```diff
@@ -170,43 +171,50 @@ module.exports = bookshelf.Model.extend(merge({
...
-    UPDATE comments SET deactivated_by_id = ${deletedUserId} WHERE deactivated_by_id = ${this.id};
     UPDATE comments SET text = 'Comment by deleted user' WHERE user_id = ${this.id};
 
-    UPDATE follows set added_by_id = ${deletedUserId} WHERE added_by_id = ${this.id};
-    UPDATE groups_tags set user_id = ${deletedUserId} WHERE user_id = ${this.id};
     DELETE FROM thanks WHERE comment_id in (select id from comments WHERE user_id = ${this.id});
     DELETE FROM thanks WHERE thanked_by_id = ${this.id};
     DELETE FROM notifications WHERE activity_id in (select id from activities WHERE reader_id = ${this.id});
     DELETE FROM notifications WHERE activity_id in (select id from activities WHERE actor_id = ${this.id});
+    DELETE FROM notifications WHERE activity_id in (select id from activities WHERE parent_comment_id in (select id from comments where user_id = ${this.id}));
```
tibetsprague, line 210: so now that im thinking about this, won't this delete notifications for other people that have comments on this person's comments? I think we need to check that the notification or the activity also is for this particular user? im a bit confused...

api/models/User.js
```diff
@@ -221,7 +229,31 @@ module.exports = bookshelf.Model.extend(merge({
     DELETE FROM posts_tags WHERE post_id in (select id from posts WHERE user_id = ${this.id});
     DELETE FROM groups_posts WHERE post_id in (select id from posts WHERE user_id = ${this.id});
     DELETE FROM votes WHERE user_id = ${this.id};
-    DELETE FROM users WHERE id = ${this.id};
+
+    UPDATE users SET 
+    active = false,
```
tibetsprague, line 234: TBD

api/models/User.js
```diff
@@ -221,7 +229,31 @@ module.exports = bookshelf.Model.extend(merge({
...
+    bio = NULL, 
+    banner_url = NULL,
+    location = NULL,
+    url = NULL,
+    tagline = NULL,
+    stripe_account_id = NULL,
+    location_id = NULL,
+    contact_email = NULL,
+    contact_phone = NULL,
+    email = '${uuid.v4()}@hylo.com',
```
tibetsprague, line 247: still thinking about this one too... I guess this looks best

thomasgwatson: Yeah, just needed a unique value

api/models/User.js
thomasgwatson: I think this was added to ensure its counterpart of

`DELETE from activities WHERE parent_comment_id in (select id from comments where user_id = ${this.id});` would work without hitting a constraint.

However, I wonder now whether this `DELETE from` is even necessary and whether they can both be removed?

api/models/User.js
thomasgwatson: What are our options here if we flip back to active, considering we would want to turn off notifications? When sending notifications, check the user name and if it matches "Deleted User", filter it out? Are there other things we'd want to skip doing?

Alternatively, if its just that we want posts and comments from deleted users to still show up, do we just change the current queries for those to now filter out non-active users?? What non-active users do we actually not want to see posts or comments for?

api/models/User.js
tibetsprague: I think we should talk this through

api/models/User.js
tibetsprague: yes it seems like the whole point is to have the posts and comments still show up. but im not sure we want them to show up for people who mark themselves as deactivated. again, lets talk it through

tibetsprague (approved):
this seems ready to test again!

PR #723, opened 2021-12-17
For deactivated/deleted user make sure their posts cant be found in t…

tibetsprague:
…he streams but still load if directly accessed


thomasgwatson (approved):
lgtm

PR #732, opened 2022-01-12
Node 718 dummy farm data seed

### LINKED ISSUE #718: Generate dummy farm data
Generate SQL that will allow any developer to rapidly add:
- several new groups with the farm type
- each of those groups having a farm extension, with a realistic payload based on the first draft of the farm profile schema

Could use knex seed file system

thomasgwatson:
**trying to avoid the wrong files coming up in the diff**

Closes #718

Generate fake farm data

Includes random gaps in the data; randomly some fields won't be filled out, giving us a feel for what comparisons we can't make and handling gaps gracefully.

Originally I was using import/export ES6 stuff but knex seed requires commonJS so there are some wrenches thrown into the works there, especially around code-reuse.

I am expecting the farm profile schema to update a little in the coming weeks, so I might add a second seed option with a specific file that refreshes the farm profile data but don't do anything with the groups or users themselves.


migrations/20220110135403_add-farm-onboarding-extension.js
```diff
@@ -0,0 +1,13 @@
+
+exports.up = function (knex) {
+  return knex.raw(`
+    INSERT INTO extensions(type, created_at) VALUES ('farm-onboarding', current_timestamp)
```
tibetsprague, line 4: extensions is still a bit confusing of a name to me, can we call it group_extensions at least?

lib/constants.js
```diff
@@ -0,0 +1,3122 @@
+'use strict'
```
tibetsprague: seems like this file should go in seeds/farm-dev/

migrations/20220110135403_add-farm-onboarding-extension.js
thomasgwatson: group_extensions is really close to the join table's name, groups_extensions. This is the same pattern as widgets and groups_widgets.

Issue #668, opened 2021-09-08
Implement login with Hylo oAuth flow


tibetsprague:
Using https://github.com/panva/node-oidc-provider

PR #733, opened 2022-01-14
Support for being an OpenID Connect provider!

tibetsprague:
Wheeee

Only really tested with the Code Authorization Code Flow with PKCE

Fixes https://github.com/Hylozoic/hylo-node/issues/668


api/services/oidc/knexAdapter.js
```diff
@@ -0,0 +1,91 @@
...
+  "DeviceCode",
+  "ClientCredentials",
+  "Client",
+  "InitialAccessToken",
+  "RegistrationAccessToken",
+  "Interaction",
+  "ReplayDetection",
+  "PushedAuthorizationRequest",
+  "Grant",
+].reduce((map, name, i) => ({ ...map, [name]: i + 1 }), {})
```
lorenjohnson, line 17: This feels a wee bit fragile given that if another type is not carefully added (at the end) later then the mapping could be shifted and I presume break things in a bad way. Maybe a comment and/or just a more explicit mapping?

config/routes.js
```diff
@@ -5,6 +5,10 @@
  */
 
 module.exports.routes = {
+  // for OIDC
+  // XXX: unclear if its ok to redirect like this, can we somehow directly call the oidc-provider discovery function here?
+  'GET    /.well-known/openid-configuration':             '/noo/oauth/.well-known/openid-configuration',
```
lorenjohnson, line 10: ...this whole thing is so nicely modular that it'd be cool if this route could somehow get snuck-into the oidc-provider routing above, but I bet you thought of that and likely it was more trouble than it's worth--if even possible.

README.md
```diff
@@ -136,6 +136,11 @@ Change your `.env` file to have:
 PROTOCOL=https
 ```
 
+### Setting up as OpenID Connect provider
+- Run `yarn generate-rsa-key-base64`
+- Copy generated base64 string to .env file: `OIDC_JWK_KEYS=base64key`
+- You can add multipke keys by separating them with a comma
```
lorenjohnson, line 142: Typo on "multipke" and curious why one would generate multiple keys? Do they get rotated over?

lorenjohnson (approved):
Overall looks good. 

1) Maybe the `types` map is a worthwhile adjustment? The rest at your discretion if something struck a cord. 

2) You have a few TODOs/open questions, which I'm fine with, and have you answered the open questions to your satisfaction? As this is basically going into "beta" testing I imagine some of those things are to be fleshed-out in testing.

lorenjohnson (approved):
I once again approve! Looks just fine to my eyeballs to the best of my understanding of the whole setup. Lemme know if there is any particular area you'd like me to consider or notice, otherwise I say "off to the races!"

PR #734, opened 2022-01-14
Add UTC to end of pretty ranges, for email purposes

### LINKED ISSUE #694: Add times + timezone to dates in email notifications
Show when an event is happening in our email notifications

When displaying a date in an email notification add the timezone to the end, like we do for events on the web. 

Do the timestamps on the server store the zone coming from the front-end? Or are they all converted to a single server time zone in the database? If the former we can display the timezone of the event, if the latter let's pick UTC and show that for all events.

thomasgwatson:
Closes #694

### DIFF
--- a/api/models/event/mixin.js
+++ b/api/models/event/mixin.js
@@ -66,7 +66,7 @@ export default {
       }
     }
 
-    return from + to + " UTC timezone"
+    return from + to + " UTC"
   },
 
   createInviteNotifications: async function (userId, inviteeIds) {
--- a/api/models/event/mixin.js
+++ b/api/models/event/mixin.js
@@ -66,7 +66,7 @@ export default {
       }
     }
 
-    return from + to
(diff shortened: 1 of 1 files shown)


api/models/event/mixin.js
```diff
@@ -66,7 +66,7 @@ export default {
       }
     }
 
-    return from + to
+    return from + to + " UTC timezone"
```
tibetsprague, line 69: is the emails the only place this is used, just want to make sure this isnt weird somewhere else? also i dont think we need "timezone", UTC is fine

tibetsprague (changes requested):

api/models/event/mixin.js
thomasgwatson: Ok. And yes, only used in the context of emails.

tibetsprague (approved):

Issue #585, opened 2021-01-13
`messageThreadId` is always null in graphql results when fetching persond


lorenjohnson:
Mobile is ready for this fix and will make it work so that an existing thread with a user will be retrieved when clicking on the "Message" button on their profile. 

Web may need to be revisited once this is fixed (see hylo-evo#748)

PR #738, opened 2022-01-20
Correctly return message thread id for a pair and current user

tibetsprague:
Fixes: https://github.com/Hylozoic/hylo-evo/issues/748 and https://github.com/Hylozoic/hylo-node/issues/585


lorenjohnson (approved):
Yay! Was hoping it was something simple like that but didn't get to that on my first pass way back whenever.

PR #739, opened 2022-01-24
Node 16 and Heroku-20 update

lorenjohnson:
This is basically free, everything seems to be happy on staging without changes 👍🏼  For testing this is what is currently deployed on hylo-node-staging as well as I've been running against node 16 in local dev on hylo-node for a couple days now and haven't seen any issues. 

* This is a nice-to-have for dev as evo, node, shortly also mobile, will converging on the same version of node. As noted on the evo ticket, there is a pre-built arm64/M1 package for Node 16 so it makes dev servers snappier to boot. Also for our new devs coming-on this slightly simplifies setup. * Annie is on an M1 as well, not sure about Sofia.

* Also, not unimportantly there is evidence that we'll have non-trivial performance gains in moving up to 16 in production as well. I looked into it a little and it seems node 14 was a bit of dog in the series, with node 16 getting back a lot of steam with the v9 engine. I would also be hopeful that moving to Heroku-20 (Ubuntu 20.04) is a plus in terms of possible performance or at least reliability. 

* Upgrades our CircleCi config to use CircleCI "Next-generation convenience images" increasing the full run CI from 4-6 mins to 1-2.5 mins. Yay! Also makes our the config a lot simpler and less opaque. Mostly the same changes can be made on hylo-evo's CircleCI setup fairly readily when we get to it.

(Moved back into draft for now as tests are failing on CircleCI for yet unknown reason)


package.json
```diff
@@ -11,7 +11,7 @@
     "url": "git://github.com/Hylozoic/hylo-node.git"
   },
   "engines": {
-    "node": "^14.17.6",
+    "node": "^16",
```
lorenjohnson, line 14: This vagueness is on purpose in the spirit of this designator evidently in this context being the "minimum version" 🤷🏼‍♂️

.circleci/config.yml
```diff
@@ -53,46 +53,47 @@ jobs:
...
-    - run: sudo apt update
-    - run: sudo apt install yarn
+    # - run: sudo sh -c 'echo "deb https://dl.yarnpkg.com/debian/ stable main" >> /etc/apt/sources.list.d/yarn.list'
+    # - run: sudo apt update
+    # - run: sudo apt install yarn
     # - run: curl -o- -L https://yarnpkg.com/install.sh | bash
     # Dependencies
-    #   This would typically go in either a build or a build-and-test job when using workflows
-    # Restore the dependency cache
-    - restore_cache:
```
tibetsprague, line 70: why remove the cache?

tibetsprague (commented):
looks good to me! curious about removing the circleCI dependency cache though which is meant to speed that up

lorenjohnson:
Oh sorry about that, I pulled it back into draft after assigning because tests were breaking in CircleCI after all. Will ping back for a final review when I have the Circle config right.

tibetsprague (approved):

PR #748, opened 2022-02-04
740 map load performance

tibetsprague:
Saved search finds last post id on creation - so it doesnt have to be passed in from the front-end

Allow for searching/filtering posts by isFulfilled, multiple types and multiple topics
Don't remove old APIs for filtering by single topic or type so as not to break mobile app

When sorting people by name lowercase their name first (for member drawer on new map)

Fixes https://github.com/Hylozoic/hylo-node/issues/740


thomasgwatson (approved):
lgtm

PR #751, opened 2022-02-10
742 normalize text handling

lorenjohnson:
From CHANGELOG:

- Move HTML sanitization exclusively of posts.details and comments.text to model getters (`Post#details()` and `Comment.text()`) which should be used EXCLUSIVELY for any output through API
- Removes extraneous and destructive HTML sanitization on plain text fields (entity decoding and protection from related React text renders)
- Adopts graphql-tools default getters for Post model to reduce duplication of code and make model results canonical for sanitization
- Move from deprecated `hylo-utils` to new `hylo-shared` library


lib/group/digest2/formatData.js
```diff
@@ -25,7 +25,7 @@ const presentPost = curry((slug, post) => {
   return pickBy(x => x, {
     id: post.id,
     title: post.get('name'),
-    details: RichText.qualifyLinks(post.get('description'), null, null, slug),
+    details: RichText.qualifyLinks(post.details && post.details(), null, null, slug),
```
tibetsprague, line 28: why check for post.details here (but not below in savedSearches)?

lib/rollbar.js
```diff
@@ -8,7 +8,6 @@ if (process.env.ROLLBAR_SERVER_TOKEN && process.env.NODE_ENV !== 'test') {
   })
   module.exports = rollbar
 } else {
-  console.log('Rollbar disabled (process.env.ROLLBAR_SERVER_TOKEN undefined)')
```
tibetsprague, line 11: thanks!

tibetsprague (approved):

lib/group/digest2/formatData.js
lorenjohnson: Probably no good reason. About to replace `RichText` in the follow-on branch so will normalize this then. probably doesn't need to be checked at all but will confirm on that round.

PR #752, opened 2022-02-12
Node 701 group explorer queries

### LINKED ISSUE #701: Query for groups near a location sorted by closeness

thomasgwatson:
Closes #701


api/graphql/schema.graphql
```diff
@@ -615,7 +615,8 @@ type Query {
     parentSlugs: [String],
     search: String,
     sortBy: String,
-    visibility: Int
+    visibility: Int,
+    coord: PointInput
```
tibetsprague, line 619: let's alphabetize

tibetsprague: also if this param always means "near this coordinate" then i wonder if we can come up with a more descriptive name. nearCoord? aroundCoord?

api/services/Search.js
```diff
@@ -28,7 +38,7 @@ module.exports = {
       }
 
       if (opts.visibility) {
-        qb.where('groups.visibility', opts.visibility)
+        qb.whereIn('groups.visibility', opts.visibility)
```
tibetsprague, line 41: does this work if opts.visibility is not an array? i want to make sure we dont break existing queries from mobile app

api/services/Search.js
```diff
@@ -45,15 +55,16 @@ module.exports = {
...
+        boundingBox: opts.boundingBox
+      }, qb)
       // this counts total rows matching the criteria, disregarding limit,
       // which is useful for pagination
       countTotal(qb, 'groups', opts.totalColumnName)
-
       qb.limit(opts.limit)
       qb.offset(opts.offset)
-      qb.groupBy('groups.id')
+      if (!opts.coord && !opts.sort === 'size') {
```
tibetsprague, line 65: a comment here would be good. even if you may not totally understand why

api/services/Search.js
thomasgwatson: There are three cases of this pattern, in forGroups, forTags and forPosts, and this was the only `where`. The other two are `whereIn`. @lorenjohnson do you recall if the mobile app has any group search queries, especially if they pass visibility? I figured we should standardize on array as input (although I think makeModels.js has a default value that I would need to change as well)

api/graphql/schema.graphql
thomasgwatson: Updated

tibetsprague (approved):

PR #757, opened 2022-02-22
Farm feature branch merged into dev

thomasgwatson:
I want to get the `farm` feature branch merged into `dev`, so that NODE work can just continue synced on 'dev', instead of on a feature branch.

Changes on the `farm` NODE branch should be deployable without impact on existing PROD env; all dev EVO features should be happy being served by a `farm` merged to `dev` backend,


api/models/Group.js
```diff
@@ -276,9 +281,10 @@ module.exports = bookshelf.Model.extend(merge({
   },
 
   createInitialWidgets: async function (transacting) {
-    const initialWidgets = await Widget.query(q => q.whereIn('id', [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])).fetchAll({ transacting })
+    // In the future this will have to look up the template of whatever group is being created and add widgets based on that
+    const initialWidgets = await Widget.query(q => q.whereIn('id', [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 19, 20, 21])).fetchAll({ transacting })
     Promise.map(initialWidgets.models, async (widget) => {
-      await GroupWidget.create({ group_id: this.id, widget_id: widget.id, order: widget.id }, { transacting })
+      await GroupWidget.create({ group_id: this.id, widget_id: widget.id, order: widget.id, context: widget.id > 10 ? 'group_profile' : 'landing' }, { transacting })
```
tibetsprague, line 287: this is a bit of a hack, but ok for now

api/models/GroupExtension.js
```diff
@@ -0,0 +1,14 @@
+module.exports = bookshelf.Model.extend(Object.assign({
+  tableName: 'group_extensions',
+  requireFetch: false,
+  hasTimestamps: true,
+  group () {
```
tibetsprague, line 5: prefer lines between function definitions

knexfile.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
   test: defaults,
   development: defaults,
   dummy: Object.assign({}, defaults, { seeds: { directory: './seeds/dummy' } }),
+  farmdev: Object.assign({}, defaults, { seeds: { directory: './seeds/farm-dev' } }),
```
tibetsprague, line 41: would love to figure out a better seeding system. why not always include a dummy farm or two in our seeds? but nothing to do for now

migrations/20210914160014_group_extensions.js
```diff
@@ -0,0 +1,32 @@
...
+    table.bigInteger('extension_id').references('id').inTable('extensions').notNullable()
+    table.jsonb('data')
+    table.boolean('active').defaultsTo(true)
+    table.index('group_id')
+    table.timestamp('created_at')
+    table.timestamp('updated_at')
+  })
+
+  await knex.schema.alterTable('groups', async (table) => {
+    table.text('type')
```
tibetsprague, line 21: hmm, isn't text a large text field without a size so shouldn't this be a string type?

migrations/20210914160014_group_extensions.js
```diff
@@ -0,0 +1,32 @@
+exports.up = async function (knex) {
+  await knex.schema.createTable('extensions', table => {
+    table.increments().primary()
+    table.text('type')
```
tibetsprague, line 4: this should a string not a unlimited size text field

migrations/20211103120345_add_new_group_widgets.js
```diff
@@ -0,0 +1,19 @@
+
+exports.up = function (knex) {
+  return knex('widgets').insert([
+    {id: 11, name: 'nearby_relevant_groups' },
```
tibetsprague, line 4: I wonder if we wait on this until we know exactly which widgets we want/need? since we are not doing farm comparison for now right?

migrations/20211201111510_add-context-column-to-group-widgets.js
```diff
@@ -0,0 +1,32 @@
...
+  await knex.schema.table('group_widgets', t => {
+    t.string('context')
+  })
+
+  const now = new Date().toISOString()
+
+  await knex.raw(`
+  INSERT INTO "public"."widgets"("id","name","created_at") VALUES
+    (15,E'opportunities_to_collaborate','${now}'),
+    (16,E'farm_map','${now}'),
```
tibetsprague, line 12: same here, hold off on new widgets until we are ready for them? but if these is easier for testing then its fine

migrations/schema.sql
```diff
@@ -828,7 +828,8 @@ CREATE TABLE public.groups (
     num_members integer,
     slack_hook_url text,
     slack_team text,
-    slack_configure_url text
+    slack_configure_url text,
+    "type" text DEFAULT 'group'::text
```
tibetsprague, line 832: ohh is it because we can only search text fields? that seems off, we really dont need a full text field for the type.

api/models/GroupExtension.js
thomasgwatson: The linter complains about this, because these function defs are in an object

migrations/20210914160014_group_extensions.js
thomasgwatson: Yeah, I didn't see this distinction. Obviously type fields don't need a big character limit. Is this worth changing?

image

migrations/20211103120345_add_new_group_widgets.js
thomasgwatson: The widget list will need to be revisited. Will include in our design chat

migrations/schema.sql
thomasgwatson: no, I just picked the wrong knex helper function

api/models/GroupExtension.js
tibetsprague: hmm, we do this everywhere though, we should probably update our linter definitions

migrations/20210914160014_group_extensions.js
tibetsprague: no performance differnence?? fascinating. ok then I guess its fine

tibetsprague: so did we decide to leave as is?

migrations/20211103120345_add_new_group_widgets.js
tibetsprague: anything to do here now or shall we merge as is and work on later?

migrations/20211201111510_add-context-column-to-group-widgets.js
thomasgwatson: I think these are all fine.

thomasgwatson:
@tibetsprague I'm down to merge this branch, however the patch test coverage is low. Not sure how much of that is because most of the patch is migrations and model setup; not a lot of business logic to test here.

tibetsprague (approved):

Issue #741, opened 2022-01-25
Make sure groups.num_members is always up to date and accurate so we can use that for memberCount


tibetsprague:
Right now we calculate memberCount every time which is  slow

- [x] Also need to update the production database to have currently accurate member counts. Migration or direct SQL update?
- [x] Create a ticket to check the database in a week or two

lorenjohnson:
Ha-- I JUST got to the "create ticket re membercount" item on my todo lists, and voila! You already got to it :)

thomasgwatson:
Calculating memberCount every-time: just checking, is there an index on that lookup? I scanned through the 40 indexes we have and none of them seemed related to groups-group-memberships? I'm assuming a index on `group_id` on the `group_memberships` table would speed up lookups quite a bit, and then the index would handle keeping the count in sync

tibetsprague:
Hmm, I don’t think we need an index on it, it’s not like we are querying by it. We just need to make sure the value in the column stays accurate 

Sent by m

PR #770, opened 2022-03-15
Update group num_members in cron job

tibetsprague:
And once again return num_members for the memberCount attribute for groups for performance.

Fixes https://github.com/Hylozoic/hylo-node/issues/741


lorenjohnson (approved):
LGTM!

PR #785, opened 2022-04-21
change geoShape type from string to JSON

Sofacosta:


tibetsprague:
I don't think you committed the right diff here

Sofacosta:
No, it's not. This is weird, I'll fix it.

Issue #670, opened 2021-09-15
Store customized group nouns for group "type" and "moderators"


thomasgwatson:
Groups might want to be called "farm, ranch, temple, coven, organization, conspiracy" etc etc.
Add a new field to groups table called type_descriptor (or think of a better name for this)
Also we may want to include type_descriptor_plural too, though not sure we will use this yet...

And a group may want to call Moderators Admins, or Stewards, or...
Add a new field to groups table called moderator_descriptor (or think of a better name for this)
Also we may want to include moderator_descriptor_plural too, though not sure we will use this yet...

Groups might want to refer to their members as "farmers, land stewards, friends, comrades" etc etc etc
But for now we are not going to work on this yet, for the future. 

We are not going to build front-end to set this for now, one can only set these fields via the API. 

Add these fields to the output for the group resolves

Also allow for setting these fields in the group creation mutations

PR #789, opened 2022-04-26
Add moderator and type descriptor fields to groups

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-node/issues/670


lorenjohnson (approved):
LGTM 👍🏼

Issue #791, opened 2022-04-27
Ability to send message to moderators of a group you are not a member of


tibetsprague:
For farm opportunities to collaborate

PR #792, opened 2022-04-27
Add mutation to send a message to all moderators of any group a perso…

tibetsprague:
…n can see

Fixes https://github.com/Hylozoic/hylo-node/issues/791


api/models/Group.js
```diff
@@ -3,6 +3,10 @@ import { clone, defaults, difference, flatten, intersection, isEmpty, map, merge
 import randomstring from 'randomstring'
 import wkx from 'wkx'
 import HasSettings from './mixins/HasSettings'
+import findOrCreateThread from './post/findOrCreateThread'
+import createComment from './comment/createComment'
```
thomasgwatson, line 7: I don't see this function being used in this PR. Is it from something else?

thomasgwatson (approved):
Lgtm, but check my comment and see if that line needs to be clipped

api/models/Group.js
tibetsprague: good catch

PR #795, opened 2022-04-28
786 more api calls for coffeeshop

tibetsprague:
- Query: groups
- Query: person
- Mutation: updateGroup


thomasgwatson (approved):
lgtm and I suspect it'd be good for me to look at the PR for the first set of Api calls.

PR #805, opened 2022-05-06
Slight tweaks to how custom group moderator and type descriptors work

tibetsprague:
Default is now null and when null we use default text that is easier to translate, instead of storing default "Moderator" or "Group" in the database
Also for type descriptor if that is null then check if type is set and use that if it is, otherwise use Group.


thomasgwatson (approved):
lgtm

PR #814, opened 2022-05-26
Add demo farms for seeding prod

thomasgwatson:


tibetsprague (approved):
feels a bit odd to do this but also understood!

PR #815, opened 2022-05-28
Add type to group input schema

thomasgwatson:


api/graphql/schema.graphql
```diff
@@ -884,6 +884,7 @@ input GroupInput {
   moderatorDescriptor: String
   moderatorDescriptorPlural: String
   name: String
+  type: String
```
tibetsprague: i dont think this is enough, we also need to whitelist this attribute in the group create and/or update method

tibetsprague (approved):

tibetsprague (approved):

PR #834, opened 2022-06-27
Node 828 custom views

### LINKED ISSUE #828: Add Custom links/views to the navigation sidebar in a group
This is for salmon nation to be able to add a link to the left navigation for “Prize Submissions”. This new view will display Projects with a certain topic in a grid view.

The general feature is to let people add custom links to the left nav. These can either just be external links or can be a "saved view" and let moderators pick a set of post types to display along with topics to filter by and choose which view to use (stream, compact list or grid). We might be able to build on the existing Saved Searches functionality in the map for this.

Have to make sure that if we use Saved Search functionality we dont send emails for new posts in these views. 
Actually probably cant use this since searches are for individuals and this is for groups. New table, similar to saved_searches.
Let's call these Custom Views, table = custom_views
Columns:
- group_id
- name
- icon
- external_link: string
- view_mode: card, list, grid (map?)
- post_types: string[]
- active_posts_only: boolean (if true hide past events and completed or past due date requests, offers....)
- is_active?

custom_view_topics
- custom_view_id
- tag_id

Questions:
- Do we have map boundary box? not for now

thomasgwatson:
Closes #828

### DIFF
--- a/api/graphql/makeModels.js
+++ b/api/graphql/makeModels.js
@@ -236,6 +236,7 @@ export default async function makeModels (userId, isAdmin, apiClient) {
       relations: [
         {activeMembers: { querySet: true }},
         {childGroups: {querySet: true}},
+        {customViews: {querySet: true}}, // Currently there is no queryset on the the actual graphQL type
         {groupRelationshipInvitesFrom: {querySet: true}},
         {groupRelationshipInvitesTo: {querySet: true}},
         {groupTags: {
@@ -435,12 +436,28 @@ export default async function makeModels (userId, isAdmin, apiClient) {
         'created_at',
         'status',
         'type',
-        'updated_at',
+        'updated_at'
       ],
       getters: {
         questionAnswers: i => i.questionAnswers().fetch()
       },
       relations: ['createdBy', 'fromGroup', 'toGroup']
+    },
+
+    CustomView: {
+      model: CustomView,
+      attributes: [
+        'group_id',
+        'is_active',
+        'search_text',
+        'icon',
+        'name',
+        'external_link',
+        'view_mode',
+        'active_posts_only',
+        'post_types',
+      ],
+      relations: ['group', 'topics']
     },
 
     Invitation: {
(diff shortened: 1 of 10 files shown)


api/graphql/makeModels.js
```diff
@@ -236,6 +236,7 @@ export default async function makeModels (userId, isAdmin, apiClient) {
       relations: [
         {activeMembers: { querySet: true }},
         {childGroups: {querySet: true}},
+        {customViews: {querySet: true}}, // Currently there is no queryset on the the actual graphQL type
```
thomasgwatson, line 239: @tibetsprague If this is a 1-to-1 relationship, it doesn't need a queryset. Should I actually put quertset false here?

tibetsprague: hmm, no I dont think you do

tibetsprague: but how is it 1:1, groups can have multiple customViews right?

api/graphql/mutations/group.js
```diff
@@ -115,7 +116,20 @@ export async function removeModerator (userId, personId, groupId, isRemoveFromGr
...
   const group = await getModeratedGroup(userId, groupId)
-  return group.update(convertGraphqlData(changes))
+  const convertedChanges = convertGraphqlData(changes)
+  /* 
+    - check if customView is in changes
+    - check if customView exists
+    - either update/create customView
+  */
+ 
+  if (group && convertedChanges.customView) {
```
tibetsprague, line 126: I feel like mostly we handle this type of extra work in the Group.update function itself instead of in the mutation, is there a reason you put this here?

tibetsprague (commented):
looks good, just one little potential change

api/graphql/makeModels.js
thomasgwatson: As far as I understood it, there was only going to be one custom view per group in the design. Is that wrong or is the idea we want to architect for an expansion of customViews in the future?

api/graphql/mutations/group.js
thomasgwatson: yep, this was just me following the chain of where things happening and stopping too early as I went along the chain! Sorry! Moving this

tibetsprague (approved):

PR #842, opened 2022-07-24
Changes needed for TipTap editor

lorenjohnson:
- Upgrades to latest HyloShared which has more tolerant sanitization in accordance with new formatting options in the new editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "hast-util-to-string": "^1.0.1",
     "hastscript": "^3.1.0",
     "he": "^1.2.0",
-    "hylo-shared": "^1.6.8",
+    "hylo-shared": "^1.8.0",
     "image-size": "^0.3.5",
     "jade": "^1.11.0",
     "jsonwebtoken": "^8.5.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3848,6 +3848,11 @@ cookies@~0.8.0:
     depd "~2.0.0"
     keygrip "~1.1.0"
 
+coordinate-parser@^1.0.7:
+  version "1.0.7"
+  resolved "https://registry.yarnpkg.com/coordinate-parser/-/coordinate-parser-1.0.7.tgz#9350485f03146af0f377dd51a0907a0ac684363e"
+  integrity sha512-pkcjigkAEjU5JsTYnuXLkRgR6T5fF/7GXR4p9vWJesy8fKwsheN8zC5d3sSvdMmWihHB4u48xWZ5mUCcgBIEpw==
+
 copy-descriptor@^0.1.0:
   version "0.1.1"
   resolved "https://registry.yarnpkg.com/copy-descriptor/-/copy-descriptor-0.1.1.tgz#676f6eb3c39997c2ee1ac3a924fd6124748f578d"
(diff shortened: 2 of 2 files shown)


api/services/RichText.js
```diff
@@ -1,54 +1,152 @@
-var Cheerio = require('cheerio')
```
lorenjohnson, line 1: Other than Link Preview related stuff, this file is the real heart of the changes in this branch. 

A lot of what is happening here was spread across a few places in the frontend and backend previously. 

Take a read through and let me know if you have any questions. 

---

NOTE:  I've moved here from `cheerio` to `jsdom` parsing and for now I'd like us to keep moving that way. It debatably (JSDOM has gotten a LOT better in recent years) is somewhat less performant than `cheerio` but allows for code that is more isomorphic (could be ran frontend as well) and keeps us familiar with standard DOM methods. 

We currently are using `cheerio` one other place here on `Hylo-node` and in the `Highlight...` search-related stuff on the frontend. It would be nice to convert those spots too... eventually.

api/services/RichText.js
```diff
@@ -1,54 +1,152 @@
...
+    el.parentNode.replaceChild(anchorElement, el)
+  })
+  convertSpansToAnchors(dom.querySelectorAll(
+    'span.topic, span.mention'
+  ))
+
+  forEach(el => {
+    const href = el.getAttribute('href')
+    if (href && !href.match(/^(https?|email):/)) {
+      el.setAttribute('href', Frontend.Route.prefix + href)
```
lorenjohnson, line 131: Re. `FrontEnd.Route` It would be really nice to merge all/most of Route/Path helper things into `hylo-shared` for keeping Web, Mobile, and backend stuff in sync and coherence.

api/services/RichText.js
```diff
@@ -1,54 +1,155 @@
...
+
+- Links will be generated with `/all` if a `groupSlug` is not passed
+
+- This same logic is handled dynamically on Web in `ClickCatcher`,
+  and on Mobile in the `HyloHTML` component.
+
 */
-export const getUserMentions = html => {
-  if (!html) return []
+export function qualifyLinks (processedHTML, groupSlug) {
```
lorenjohnson, line 108: This is used for generating links from  in the HTML content for emails. This is the only place in the backend which will generate URLs scoped to a group. Currently everywhere this is ran on the backend a `groupSlug` is provided and pretty much always it's first group on the Post... as we do on both front-ends.

api/services/RichText.js
```diff
@@ -1,54 +1,155 @@
...
+
+    el.parentNode.replaceChild(anchorElement, el)
+  })
+  convertSpansToAnchors(dom.querySelectorAll(
+    'span.topic, span.mention'
+  ))
+
+  forEach(el => {
+    const href = el.getAttribute('href')
+    if (href && !href.match(/^(https?|email):/)) {
```
lorenjohnson, line 133: It's almost certainly a bug to be `Frontend.Route.prefix`'ing an `email:` URL 🤔 Tough obscured by the changes, this part was this way before. Can you seen any reason this would work? Shouldn't we just leave `email:` links alone?

tibetsprague (approved):
looks good!

tibetsprague:
tests failing though

lorenjohnson:
That's because of the sql schema. I need assistance with that 😔

lorenjohnson:
> That's because of the sql schema. I need assistance with that 😔

It's actually just one test, and should pass once I get a valid schema dump

tibetsprague:
happy to help with the schema

lorenjohnson:
> happy to help with the schema

Wanna just pull it down and see if you have the magic? I tried several ways at it, which I have had to do in the past, but this time I was foiled.

PR #847, opened 2022-07-28
Allow for passing in geo_shape geoJSON when creating a group

tibetsprague:


thomasgwatson:
So, unfortunately, `geomFromGeoJSON` in postgis only accepts a single geoJSON geometry, not a feature collection. I think this was part of what was bogging us down in the prior attempt to get this sorted.

We could check if the incoming geoJSON is a feature collection, and then port it to a `wkt` encoded 'geometry collection', which can then be inserted into the `geo_shape` row via `ST_AsText`.

We will also need to validate that the same geometries are being served out as valid geoJSON. Perhaps helper functions for these sorts of operations were already built out when Sophia did the group geometry stuff?

PR #881, opened 2022-10-31
Add default general topic to all new groups

### LINKED ISSUE #874: New groups have a default #general topic always added
When a group is created add a #general GroupTag to it with is_default set to true. Also set visibility to pinned? Yes maybe, TBD.

thomasgwatson:
closes #874


lorenjohnson (approved):
LGTM

PR #892, opened 2022-11-16
870 chat rooms

tibetsprague:
really confused why this is showing 5.0.2 stuff which i already merged into the base branch 🤷‍♂️


api/models/Post.js
```diff
@@ -1,8 +1,11 @@
 /* globals _ */
+
 import data from '@emoji-mart/data'
 import { init, getEmojiDataFromNative } from 'emoji-mart'
+import insane from 'insane'
```
thomasgwatson, line 5: interesting library name

lib/graphql-bookshelf-bridge/util/applyPagination.js
```diff
@@ -7,16 +7,12 @@ export const PAGINATION_TOTAL_COLUMN_NAME = '__total'
...
-    throw new GraphQLYogaError('Specifying both cursor and offset is not supported.')
-  }
-
   if (cursor && sortBy !== 'id') {
     throw new GraphQLYogaError('Specifying both cursor and sortBy is not supported.')
   }
 
   // skip special sorts
-  if (!['join', 'votes', 'updated', 'created'].includes(sortBy)) { // Need to change this to reactions once Mobile has been ported
+  if (!['join', 'reactions', 'votes', 'updated', 'created'].includes(sortBy)) { // Need to change this to reactions once Mobile has been ported
```
thomasgwatson, line 15: I guess with this change that we can remove the comment

thomasgwatson (approved):
lgtm.

If we are changing the sortBy (from votes to reactions), does that mean this PR won't support mobile until its ported over?

tibetsprague:
> lgtm.
> 
> If we are changing the sortBy (from votes to reactions), does that mean this PR won't support mobile until its ported over?

oh gosh, yes thats probably true, we probably need to support both options for now

api/models/Post.js
lorenjohnson: This actually shouldn't be being imported here, and doesn't appear to be used anywhere in the file. For keeping our application of sanitization consistent, and not double-sanitizing which causes problems, should be done through using one of the `HyloShared.TextHelpers` methods.

lorenjohnson (approved):
Other than removing the extraneous `insane` import in the Post model this looks ready to go.

tibetsprague:
> lgtm.
> 
> If we are changing the sortBy (from votes to reactions), does that mean this PR won't support mobile until its ported over?

actually I left in all the votes sorting stuff, so hopefully still works

PR #900, opened 2022-11-30
Update comment url for notifications

### LINKED ISSUE #883: Improve notification link URLs for comment notifications
- [ ] Clicking on a comment notification link (in app and in email) should open the post and scroll to that comment
- [x] Same with comment mention notifications
- [x] Comment should be highlighted to make it stand out

thomasgwatson:
closes #883 

I want to test this in staging

### DIFF
--- a/api/services/Frontend.js
+++ b/api/services/Frontend.js
@@ -67,15 +67,13 @@ module.exports = {
 
     root: () => url('/app'),
 
-    comment: function (comment, group) {
-      // TODO: update to use comment specific url when implemented in frontend
-      let groupSlug = getSlug(group)
+    comment: function ({ comment, groupSlug, post }) {
 
       let groupUrl = isEmpty(groupSlug) ? '/all' : `/groups/${groupSlug}`
 
-      const postId = comment.relations.post.id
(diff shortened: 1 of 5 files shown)


api/services/Frontend.js
```diff
@@ -67,15 +67,13 @@ module.exports = {
...
 
-    comment: function (comment, group) {
-      // TODO: update to use comment specific url when implemented in frontend
-      let groupSlug = getSlug(group)
+    comment: function ({ comment, groupSlug, post }) {
 
       let groupUrl = isEmpty(groupSlug) ? '/all' : `/groups/${groupSlug}`
 
-      const postId = comment.relations.post.id
+      const postId = comment?.relations?.post?.id || post.id
```
tibetsprague, line 74: why would we pass the post in? cant we always get it from the comment? seems dangerous you could pass in a different post than the one the comment is on

thomasgwatson: Previously there was only one call site for this function, now three. One of the new call-sites was passing in a comment model that didn't have a post on it. So decided to pass in the post in that instance.

PR #905, opened 2022-12-11
Evo 1331 child posts

thomasgwatson:
For https://github.com/Hylozoic/hylo-evo/issues/1331


tibetsprague (approved):
I guess the question I still have is whether this one setting also affects the map? or is that a different setting? or does the map setting not get saved but get reset every time?

thomasgwatson:
It does save in the user settings on the map as well.

PR #914, opened 2023-01-11
Fixing public groups and hopefully sessions

tibetsprague:
We decided to let all new groups back into the public context, for now. 

And this upgrades old deprecated node-uuid package to newer uuid as a shot in the dark hope that the reason a couple folks have gotten logged in as others is due to duplicate uuid sessions being generated? Also probably good for security to not be using deprecated packages


thomasgwatson (approved):
lgtm

PR #915, opened 2023-01-12
Add full comment url to comment notifications

### LINKED ISSUE #907: Make sure latest comment appears in URL in email notification
When clicking through to a post from a notification email for a new comment make sure to jump to the latest comment (the one from the email)

thomasgwatson:
Most of the comment notification changes had already been made, and the one remaining one is for mobile notifications. I can't test those directly in my local dev however...

Closes #907

### DIFF
--- a/api/models/Notification.js
+++ b/api/models/Notification.js
@@ -5,6 +5,7 @@ import decode from 'ent/decode'
 import { refineOne } from './util/relations'
 import rollbar from '../../lib/rollbar'
 import { broadcast, userRoom } from '../services/Websockets'
+import { getSlug } from '../services/Frontend'
 
 const TYPE = {
   Mention: 'mention', // you are mentioned in a post or comment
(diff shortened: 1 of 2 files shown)


api/models/Notification.js
```diff
@@ -178,9 +179,12 @@ module.exports = bookshelf.Model.extend({
   },
 
   sendCommentPush: function (version) {
-    var comment = this.comment()
-    var path = url.parse(Frontend.Route.post(comment.relations.post)).path
-    var alertText = PushNotification.textForComment(comment, version)
+    const comment = this.comment()
+    const post = comment.relations.post
+    const group = post.relations.groups.first()
```
tibetsprague, line 184: this is where we would ideally figure out the best group to send the person to, though not sure its setup to customize the notification text per user just yet, so that can wait.

tibetsprague (approved):
you can test by finding the push_notification in the database and going to that path

api/models/Notification.js
thomasgwatson: Yeah, Loren and I had a deep dive into this. We determined the listing of groups when the post is created is maintained in the DB, so first is as good as any. Of course, we might have users that are seeing the post because of their membership to another group. Right now, they'll be redirected to `/all` on the front-end but yes, we could check their memberships and match that to the groups on a post as a slightly more relevant option

PR #924, opened 2023-01-24
Node 920 roles badges

### LINKED ISSUE #920: Table/model setup for roles/badges
new table group_roles
- group_id - ID
- name - string
- emoji - string
- color - string?

new table member_roles
- group_id 
- user_id (index on pair of group_id & user_id)
- group_role_id

mutations to addGroupRole, removeGroupRole (also delete's all memberRoles? or just deactivates and hides?), addRoleToMember, removeRoleFromMember...

thomasgwatson:
closes #920


api/graphql/mutations/role.js
```diff
@@ -0,0 +1,88 @@
+const { GraphQLYogaError } = require('@graphql-yoga/node')
+
+export async function addGroupRole ({ groupId, color, name, emoji, userId }){
+  if (!userId) throw new GraphQLYogaError(`No userId passed into function`)
+
+  if (groupId && name && emoji) {
```
tibetsprague, line 6: I dont think emoji should be a requirement.... maybe? lets double check with Aaron

api/graphql/mutations/role.js
```diff
@@ -0,0 +1,88 @@
...
+    const groupMembership = await GroupMembership.forIds(userId, groupId).fetch()
+
+    if (groupMembership && (groupMembership.get('role') === GroupMembership.Role.MODERATOR)) {
+      const newGroupRole = { group_id: groupId, name, emoji, active: true }
+      if (color) newGroupRole.color = color
```
tibetsprague, line 11: I think you can just include color in the object above and if its null it will be ignored

api/graphql/mutations/role.test.js
```diff
@@ -0,0 +1,67 @@
+import { expect } from 'chai'
```
tibetsprague, line 1: yay tests thank you!

api/graphql/schema.graphql
```diff
@@ -930,6 +930,17 @@ type GroupRelationshipInviteQuerySet {
   items: [GroupRelationshipInvite]
 }
 
+# Group role
+type GroupRole {
+  name: String
```
tibetsprague, line 935: alphabetize please

api/graphql/schema.graphql
```diff
@@ -1853,6 +1885,11 @@ type Mutation {
   addPeopleToProjectRole(peopleIds: [ID], projectRoleId: ID): GenericResult
   # Add a Post to a Collection
   addPostToCollection(collectionId: ID, postId: ID): GenericResult
+  # For a moderator to add a badge/role to a group
```
tibetsprague, line 1888: to a member of a group

api/graphql/schema.graphql
```diff
@@ -1990,6 +2027,8 @@ type Mutation {
   removePost(postId: ID, slug: String, groupId: ID): GenericResult
   # Remove a Post from a Collection, if you have permission to
   removePostFromCollection(collectionId: ID, postId: ID): GenericResult
+  # Remove a badge/role from a member of a group, inititated by that member or by a group moderator
```
tibetsprague, line 2030: oh interesting, the member can remove themselves from a role? I guess that makes sense but worth checking on in product meeting

api/models/MemberRole.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable camelcase */
+
+module.exports = bookshelf.Model.extend({
+  tableName: 'members_roles',
+  requireFetch: false,
```
tibetsprague, line 5: realizing that I think it would be good to add created_at and updated_at to this table to track when roles get assigned or changed

tibetsprague: thought of it here because its missing the timestamps property

api/graphql/mutations/role.js
thomasgwatson: If there is no emoji... how would anyone know someone has a badge/role? It'd be a silent role I guess

tibetsprague: I thought we would always display the name/text too

tibetsprague: except maybe in small spaces. a question for @brodeur

api/graphql/schema.graphql
thomasgwatson: I wanted to leave the option open but we don't need to do the UI for it on the first pass

api/models/MemberRole.js
thomasgwatson: Same for GroupRole ?

tibetsprague: Doesn’t feel as important but I guess if you are adding then why not

tibetsprague (approved):

PR #935, opened 2023-02-20
Node 920 roles badges

### LINKED ISSUE #920: Table/model setup for roles/badges

thomasgwatson:
Fully closes #920


api/graphql/index.js
```diff
@@ -294,7 +294,7 @@ export function makeMutations (expressContext, userId, isAdmin, fetchOne) {
     addPostToCollection: (root, { collectionId, postId }) =>
       addPostToCollection(userId, collectionId, postId),
 
-    addRoleToMember: (root, { personId, groupRoleId, groupId }) => addRoleToMember({ personId, groupRoleId, groupId }),
+    addRoleToMember: (root, { personId, groupRoleId, groupId }) => addRoleToMember({ userId, personId, groupRoleId, groupId }),
```
tibetsprague, line 297: why do we have userId and personId?

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
   }
 }
 
-export async function removeRoleFromMember ({userId, memberRoleId, personId, groupId}){
+export async function removeRoleFromMember ({userId, groupRoleId, personId, groupId}){
```
tibetsprague, line 64: userId and personId?

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
...
   if (!userId) throw new GraphQLYogaError(`No userId passed into function`)
 
-  if (personId && memberRoleId && groupId) {
+  if (personId && groupRoleId && groupId) {
```
tibetsprague, line 67: personId?

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
...
     const groupMembership = await GroupMembership.forIds(userId, groupId).fetch()
```
tibetsprague, line 68: userId?

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
...
 
     if (groupMembership && (groupMembership.get('role') === GroupMembership.Role.MODERATOR) || userId === personId) {
```
tibetsprague, line 70: userId and personId?

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
...
-      const memberRole = await MemberRole.query('where', 'id', '=', memberRoleId).fetch()
+      const memberRole = await MemberRole
+        .query(q => {
+          return q.where('user_id', personId)
```
tibetsprague, line 73: personId

api/graphql/mutations/role.js
```diff
@@ -67,20 +61,25 @@ export async function addRoleToMember ({userId, groupRoleId, personId, groupId})
...
+                  .andWhere('group_role_id', groupRoleId)
+        })
+        .fetch()
       return memberRole.destroy()
     } else {
       throw new GraphQLYogaError(`User doesn't have required privileges to remove role from member`)
     }
   } else {
-    throw new GraphQLYogaError(`Invalid/undefined parameters to remove role from member: received ${JSON.stringify({ personId, memberRoleId })}`)
+    throw new GraphQLYogaError(`Invalid/undefined parameters to remove role from member: received ${JSON.stringify({ personId, groupRoleId, groupId })}`)
```
tibetsprague, line 82: personId

api/graphql/mutations/role.js
tibetsprague, line 68: oh is userId the one assigning the badge? then i think it should be called creatorId, or moderatorId

api/graphql/mutations/role.js
tibetsprague: so this is prepping for a time when non moderators might be able to give people badges? i dont know if we really know if/what that use case might be, so not sure we need this. but open to it

api/graphql/schema.graphql
```diff
@@ -1453,6 +1451,9 @@ type Person {
     # Determines sort order, either 'ASC' or 'DESC'. Memberships always sorted by order they were created.
     order: String
   ): [Membership]
+  # All the member roles for a user
```
tibetsprague, line 1454: I would say all the group roles instead of member roles

api/graphql/schema.graphql
```diff
@@ -2032,7 +2035,7 @@ type Mutation {
   # Remove a Post from a Collection, if you have permission to
   removePostFromCollection(collectionId: ID, postId: ID): GenericResult
   # Remove a badge/role from a member of a group, inititated by that member or by a group moderator
```
tibetsprague, line 2037: oh interesting, you are saying someone can remove themself from a role? i'm actually not sure about that, if roles and badges are assigned by moderators they probably shouldnt be able to be removed by the person, only by moderators. a product question for the group

api/models/User.js
```diff
@@ -277,6 +283,7 @@ module.exports = bookshelf.Model.extend(merge({
 
     DELETE FROM thanks WHERE comment_id in (select id from comments WHERE user_id = ${this.id});
     DELETE FROM thanks WHERE thanked_by_id = ${this.id};
+    DELETE FROM members_roles WHERE user_id = ${this.id};
```
tibetsprague, line 286: good catch

api/graphql/index.js
thomasgwatson: Yeah, as you got to, userId is needed to check if they are a moderator. I was just passing userId along the whole way, because that is what it was originally called. Could switch to moderatorId

api/graphql/mutations/role.js
thomasgwatson: I just added the option for any user to be able to remove a badge from themselves in the API but there isn't that option int he front-end for now.

api/graphql/schema.graphql
thomasgwatson: Yeah. Currently its only implemented in the API, not the front-end

tibetsprague:
maybe just add a comment saying that personId is the member getting the badge/role and then this is good

tibetsprague (approved):

PR #936, opened 2023-02-24
Minor farm list updates

thomasgwatson:
https://github.com/Hylozoic/hylo-evo/issues/1462

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -447,6 +447,14 @@ exports.PREFERRED_CONTACT_METHODS = [
 
 exports.PRODUCT_CATAGORIES = [
   {
+    value: 'grains',
+    label: 'Grains'
+  },
+  {
+    value: 'other_row_crops',
+    label: 'Other row crops'
+  },
+  {
     value: 'grains_other',
     label: 'Grains and other row crops'
   },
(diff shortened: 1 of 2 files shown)


lib/constants.js
```diff
@@ -446,6 +446,14 @@ exports.PREFERRED_CONTACT_METHODS = [
...
+  {
+    value: 'grains',
+    label: 'Grains'
+  },
+  {
+    value: 'other_row_crops',
+    label: 'Other row crops'
+  },
   {
     value: 'grains_other',
```
tibetsprague, line 458: so this doesnt need to appear in the schema?

lib/constants.js
```diff
@@ -446,6 +446,14 @@ exports.PREFERRED_CONTACT_METHODS = [
 ]
 
 exports.PRODUCT_CATAGORIES = [
+  {
+    value: 'grains',
+    label: 'Grains'
+  },
+  {
+    value: 'other_row_crops',
```
tibetsprague, line 454: this doesnt need to appear in the schema?

lib/constants.js
thomasgwatson: Notionally no, but leaving it on makes this backwards compatible with existing farm groups.

lib/constants.js
thomasgwatson: Notionally no, but leaving it on makes this backwards compatible with existing farm groups.

tibetsprague (approved):

PR #948, opened 2023-04-28
Node 888 i18n

### LINKED ISSUE #888: Internationalization of server strings
- [ ] Emails in SendWithUs
- [ ] Notifications
- [ ] Error messages? Should probably switch to error codes and have the messages on the front-end. For now just translate what we have on the server
- [ ] Anything else?

thomasgwatson:
Closes #888


thomasgwatson:
eugh, tests are failing here but passing locally :(

api/models/Group.js
```diff
@@ -10,12 +10,14 @@ import HasSettings from './mixins/HasSettings'
 import findOrCreateThread from './post/findOrCreateThread'
 import { groupFilter } from '../graphql/filters'
 import { inviteGroupToGroup } from '../graphql/mutations/group.js'
-
 import DataType, {
   getDataTypeForInstance, getDataTypeForModel, getModelForDataType
 } from './group/DataType'
 import convertGraphqlData from '../graphql/mutations/convertGraphqlData'
 import { findOrCreateLocation } from '../graphql/mutations/location'
+import { es } from '../../lib/i18n/es'
```
tibetsprague, line 18: this is what should ultimately be in hylo-shared no? but ok for now to have it here

api/models/Group.js
```diff
@@ -750,15 +752,16 @@ module.exports = bookshelf.Model.extend(merge({
...
+      const locale = creator.get('settings')?.locale || 'en'
       return Email.sendRawEmail(recipient, {
-        subject: "New Hylo Group Created: " + g.get('name'),
-        body: `Group
-          Name: ${g.get('name')}
+        subject: locales[locale].groupCreatedNotifySubject(g.get('name')),
+        body: `${locales[locale].Group()}
+          ${locales[locale].Name()}: ${g.get('name')}
           URL: ${Frontend.Route.group(g)}
-          Creator Email: ${creator.get('email')}
```
tibetsprague, line 758: this is an email that goes to us as Hylo stewards telling us about new groups, it really doesnt need to be translated. but this is ok

api/models/Notification.js
```diff
@@ -134,46 +134,50 @@ module.exports = bookshelf.Model.extend({
   sendEventInvitationPush: function () {
     const post = this.post()
     const actor = this.actor()
+    const locale = this.reader().get('settings')?.locale || this.actor().get('settings')?.locale || 'en'
```
tibetsprague, line 137: I think it should always be the reader, not the actor right? though I guess that could be a way to guess if we dont have the reader's locale

api/models/Notification.js
```diff
@@ -134,46 +134,50 @@ module.exports = bookshelf.Model.extend({
...
-        var alertText = PushNotification.textForAnnouncement(post)
+        var alertText = PushNotification.textForAnnouncement(post, locale)
         return this.reader().sendPushNotification(alertText, path)
       })
   },
 
   sendPostPush: function (version) {
     var post = this.post()
     var groupIds = Activity.groupIds(this.relations.activity)
+    const locale = this.reader().get('settings')?.locale || this.actor().get('settings')?.locale || 'en'
```
tibetsprague, line 164: seems like this should be turned into a helper function

api/models/Notification.js
```diff
@@ -506,10 +531,13 @@ module.exports = bookshelf.Model.extend({
     const reader = this.reader()
     const childGroup = await this.relations.activity.otherGroup().fetch()
     const parentGroup = await this.relations.activity.group().fetch()
+    const locale = this.reader().get('settings')?.locale || this.actor().get('settings')?.locale || 'en'
```
tibetsprague, line 534: definitely a helper function (getter) for this

api/models/comment/notifications.js
```diff
@@ -65,6 +66,7 @@ export const sendDigests = async () => {
       // read time.
       let lastReadAt = user.pivot.get('last_read_at')
       if (lastReadAt) lastReadAt = new Date(lastReadAt)
+      const locale = mapLocaleToSendWithUS(user.get('settings').locale || 'en-US')
```
tibetsprague, line 69: why en-US here but just en elsewhere?

lib/i18n/en.js
```diff
@@ -0,0 +1,39 @@
...
+  textForGroupParentGroupJoinRequestAcceptedChildModerator: ({actor, parentGroup, childGroup}) => `Your group ${childGroup.name} has been accepted as a member of ${parentGroup.name} by ${actor.get('name')}`,
+  textForGroupParentGroupJoinRequestAcceptedParentMember: ({parentGroup, childGroup}) => `The group ${childGroup.name} just joined your group ${parentGroup.name}!`,
+  textForGroupParentGroupJoinRequestAcceptedChildMember: ({parentGroup, childGroup}) => `Your group ${childGroup.name} has joined ${parentGroup.name}.`,
+  textForDonationTo: ({amount, postName}) => `You contributed $${amount} to "${postName}"`,
+  textForDonationFrom: ({amount, actor, postName}) => `${actor.get('name')} contributed $${amount} to "${postName}"`,
+  newSavedSearchResults: (name) => `New saved search results in ${name}`,
+  recentActivityFrom: (name) => `Recent activity from ${name}`,
+  createInvitationSubject: (name) => `Join me in ${name} on Hylo!`,
+  apiInviteMessageContent: (name) => `${name} is excited to invite you to join our community on Hylo.`,
```
tibetsprague, line 29: I think a good idea to alphabetize this whole file

lib/util.js
```diff
@@ -17,6 +17,17 @@ module.exports = {
 
   socketIo: function () {
     return socketIoEmitter(RedisClient.create())
+  },
+
+  mapLocaleToSendWithUS: function (locale) {
+    const lookup = {
+      'en': 'en-US',
```
tibetsprague, line 24: I see this here, but then that makes it confusing when you pass in en-US up above, since it will do the mapping here

api/models/Group.js
thomasgwatson: These are only server-side strings, and they don't use the same convention as evo. I don't see any value in having these strings in shared.

There might be overlap of strings between evo and mobile, and they should be synced up on using the same framework, so I can see the value in them sharing something in shared.

api/models/Notification.js
thomasgwatson: Discussion for this here: https://www.hylo.com/my/posts/post/60053

api/models/Notification.js
thomasgwatson: ok

api/models/comment/notifications.js
thomasgwatson: sendwithus need FULL locale, so we need to convert before sending

lib/util.js
thomasgwatson: `mapLocaleToSendWithUS(user.get('settings').locale || 'en-US')` the || `en-US` is in case the locale is undefined out of settings

api/models/Notification.js
tibetsprague: ah hmm, ok this is fine for now I guess 🤷‍♂️

tibetsprague (approved):

api/models/Notification.js
thomasgwatson: Do we want give all users a default locale?

tibetsprague:
Honestly we probably should for safety in case we forget to add a default somewhere, start with ‘en’ of course

PR #952, opened 2023-07-12
Only show pinned posts from current (parent) group, not child groups

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/1479

Also some other small things

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -44,6 +44,7 @@ module.exports = {
       "https://hylo.com",
       "https://api.hylo.com",
       "https://staging.hylo.com",
+      "https://node1.hylo.com",
       "https://review.hylo.com",
       "https://api-staging.hylo.com",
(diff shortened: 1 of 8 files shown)


config/env/production.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
       "https://hylo.com",
       "https://api.hylo.com",
       "https://staging.hylo.com",
+      "https://node1.hylo.com",
```
thomasgwatson, line 47: is this a node backend replicant?

thomasgwatson (approved):
lgtm

config/env/production.js
tibetsprague: its one of the urls we use with heroku for our backend yes. I'm not actually sure we need this, but was running into a weird issue accessing the backend for some API stuff and thought this might fix it. it's safe to do at least

Issue #939, opened 2023-03-14
Time zone appearing in emails is wrong


tibetsprague:
Im seeing the time on events as 7 hours off from my timezone

PR #957, opened 2023-08-09
939 timezone in emails

tibetsprague:
Show correct timezone for events in emails

For now this always shows the timezone set by/when the person created the event as the event's timezone, and displays that in the email.

Fixes https://github.com/Hylozoic/hylo-node/issues/939


thomasgwatson (approved):
lgtm

PR #967, opened 2023-10-12
Add chat urls to notifications

thomasgwatson:
For https://github.com/Hylozoic/HyloReactNative/issues/663

### DIFF
--- a/api/models/Post.js
+++ b/api/models/Post.js
@@ -707,7 +707,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
             title: post.summary(),
             type: post.get('type'),
             url: Frontend.Route.post(post),
-            groups: post.relations.groups.map(g => ({ id: g.id, name: g.get('name'), url: Frontend.Route.group(g), postUrl: Frontend.Route.post(post, g) })),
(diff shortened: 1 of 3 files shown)


api/models/Post.js
```diff
@@ -707,7 +707,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
             title: post.summary(),
             type: post.get('type'),
             url: Frontend.Route.post(post),
-            groups: post.relations.groups.map(g => ({ id: g.id, name: g.get('name'), url: Frontend.Route.group(g), postUrl: Frontend.Route.post(post, g) })),
+            groups: post.relations.groups.map(g => ({ id: g.id, name: g.get('name'), url: Frontend.Route.group(g), postUrl: Frontend.Route.post(post, g) })), // This is the thing that is probably borking the zapier links
```
tibetsprague, line 710: I have a fix for the zapper links ready to go

tibetsprague (approved):

api/models/Post.js
thomasgwatson: ok, yeah I was just 'in the area'... I guess if folks are getting URLs from zapier about chat posts they are going to be still be weird from a UX perspective... Do we want to fix those in the same way as the other chat post notifications now?

tibetsprague: Yes probably

PR #968, opened 2023-11-04
Odds and ends

tibetsprague:
- Turn off allow in public for new groups by default
- Fix null post time for notification emails for posts that dont have a start time
- Don't sort pinned posts first when getting posts for chat rooms
- Improve performance by loading eventInvites only for events and post members only for projects
- For post zapier triggers set URL to be post in URL in the first group instead of in the "all" groups context
- Cleanup


thomasgwatson (approved):
lgtm

PR #974, opened 2024-01-19
973-join-questions-on-invite

tibetsprague:
Enable asking join questions for open groups and closed groups

Rename JoinRequestQuestionAnswer to GroupJoinQuestionAnswer


api/models/Group.js
```diff
@@ -471,7 +472,7 @@ module.exports = bookshelf.Model.extend(merge({
           // Make sure that the user making the changes doesn't need to then accept the new agreements
           const updatedByUserMembership = await GroupMembership.forPair(updatedByUserId, this.id).fetch()
           if (updatedByUserMembership) {
-            await updatedByUserMembership.save({ settings: { agreementsAcceptedAt: (new Date()).toISOString() } }, { transacting })
+            await updatedByUserMembership.save({ settings: Object.assign(updatedByUserMembership.settings, { agreementsAcceptedAt: (new Date()).toISOString() }) }, { transacting })
```
thomasgwatson, line 475: Any particular reason to use Object.assign instead of the spread operator here?

api/models/GroupJoinQuestionAnswer.js
```diff
@@ -1,16 +1,24 @@
 module.exports = bookshelf.Model.extend(Object.assign({
-  tableName: 'join_request_question_answers',
+  tableName: 'group_join_questions_answers',
```
thomasgwatson, line 2: Been pondering about how we do join tables, since its a thing for responsibilities...

Should we defaulting to loading in user and group ids to these tables, to simplify lookups or SQL queries?

api/models/Group.js
tibetsprague: in the moment i wasnt getting spread to work there 🤷‍♂️

thomasgwatson (approved):
lgtm

api/models/GroupJoinQuestionAnswer.js
tibetsprague: thats what i did in this case, as you can see the migration. i think it depends on the use case

tibetsprague: in this case we need group and user because we wont always have a jon request anymore

PR #978, opened 2024-01-31
Fix bug that allowed people to message someone they don't share a group with

tibetsprague:


api/models/post/findOrCreateThread.js
```diff
@@ -1,40 +1,41 @@
...
   if (!(participantIds && participantIds.length)) {
     throw new GraphQLYogaError("participantIds can't be empty")
   }
-  const checkForSharedGroup = id =>
-    Group.inSameGroup([userId, id])
-    .then(doesShare => {
-      if (!doesShare) throw new GraphQLYogaError(`no shared groups with user ${id}`)
-    })
-  return Promise.all(map(checkForSharedGroup, participantIds))
+  const validParticipantIds = await personFilter(userId)(User.where('id', 'in', uniq(participantIds))).fetchAll()
```
thomasgwatson, line 34: Does this mean that everyone in a group chat needs to share a group?

thomasgwatson: Maybe just the thread creator (looks at the test descriptions)

thomasgwatson (approved):
lgtm

tibetsprague:
It means that the person creating the thread needs to share a group with everyone in the thread. On Jan 31, 2024, at 5:42 PM, Tom Watson ***@***.***> wrote:﻿
@thomasgwatson commented on this pull request.



In api/models/post/findOrCreateThread.js:
>    if (!(participantIds && participantIds.length)) {
     throw new GraphQLYogaError("participantIds can't be empty")
   }
-  const checkForSharedGroup = id =>
-    Group.inSameGroup([userId, id])
-    .then(doesShare => {
-      if (!doesShare) throw new GraphQLYogaError(`no shared groups with user ${id}`)
-    })
-  return Promise.all(map(checkForSharedGroup, participantIds))
+  const validParticipantIds = await personFilter(userId)(User.where('id', 'in', uniq(participantIds))).fetchAll()

Does this mean that everyone in a group chat needs to share a group?

Issue #832, opened 2022-06-22
Add option to receive an email/push notification for every post in a group


tibetsprague:
When this is on i think there's a very small tweak in Activity.js #generateNotificationMedia where it says
// Send emails if email notifications on and is announcement or "not just a new post" notification
here we just turn off the check for !isJustNewPost and then an email should get sent for every post in the group, I think?

Is there an issue in the fact that digests are only all or nothing and so if you turn this on for a group you would get an email for every post and also a digest? should digest also be controllable on a group by group basis?

Current state of notification controls (see Activity.generateNotificationMedia, Post.createActivities, Activity.saveForReasons)
User wide:
- Digest frequency = daily, weekly or none (this covers saved searches too)
- DM notifications by push, email or both
- Comment notifications by push, email or both

Per Group 
- Send Email = whether to get emails for digests. for posts in topics followed? announcements? post mentions? comment mentions? event invitations?
- Send push notifications = all new posts? announcements? posts in topics? post mentions? comment mentions?

tibetsprague:
after further discussion we agreed that digests should also now be controllable on a group by group basis

PR #980, opened 2024-02-16
Allow for option to get notifications for every post

tibetsprague:
Send an in-app notification always for every post. Improvements to notification text.
Option to only get "important" post notifications which is announcements and mentions, or turn off post notifications completely.

Fixes https://github.com/Hylozoic/hylo-node/issues/832


api/models/Activity.js
```diff
@@ -264,35 +274,40 @@ module.exports = bookshelf.Model.extend({
     if (!isJoinRequestRelated) await activity.load('post.groups')
 
     // TODO: rename 'notifications' to 'media'
```
thomasgwatson, line 276: is this still relevant?

tibetsprague:
Hmm, that is very old. No?

thomasgwatson (approved):
lgtm

PR #984, opened 2024-02-27
Proposals

### LINKED ISSUE #969: Proposal post type
https://docs.google.com/document/d/1EUiiTtIyGAIEZRwF1uJtDk-FAyj3HchRZ5Kwf62cSv8/edit#heading=h.j8sxm8kc5x8

- [ ] New post type 'proposal'
- - [ ] New fields: Quorum = minimum # votes needed for vote to pass
- - [ ] Status: options = Discussion phase , voting phase, completed, quorum not reached (or is this an outcome?). The status changes from discussion to voting when start_time starts, or can be done manually. It changes from voting to completed when end_time reached if that is set, or can be changed manually.
- - [ ] Outcome: Pass, Fail, Quorum not Reached (or is this a Status?)
- - [ ] Anonymous? boolean (this is for v2, unless its super easy to squeeze in)
- [ ] proposal_options table
- - [ ] proposal_id
- - [ ] text: string (required. meant to be a short option like "yes", "consent"...)
- - [ ] description: text (optional. to describe what voting this option really means)
- [ ] proposal_votes table
- - [ ] proposal_id
- - [ ] option_id
- - [ ] user_id
- - [ ] created_at
- - [ ] updated_at (can people change their votes?? i think so)
- bonus: a new responsibility which allows you to create proposals called Governance (and a new common role for it? or maybe it just one people can use for their own roles.) By default everyone can do this?

thomasgwatson:
Closes #969


thomasgwatson:
Lots of `api/graphql/index.test.js` tests are broken... I feel like there is a common culprit for this (its happened in the past). Will dig into it in the morning

thomasgwatson:
Was mutations that were in the resolvers but not in the schema, fixed. Good to review now

api/models/Post.js
```diff
@@ -93,7 +97,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
     return this.get('num_people_reacts')
   },
 
-  votesTotal: function () {
+  votesTotal: function () { // TODO PROPOSAL: check if this can be removed
```
tibetsprague, line 100: flagging TODO

api/models/Post.js
```diff
@@ -201,11 +213,11 @@ module.exports = bookshelf.Model.extend(Object.assign({
 
     return q
   },
-
+  // TOOD PROPOSAL: check if this can be removed
```
tibetsprague, line 216: i wont flag all of them, but there are more

api/models/Post.js
```diff
@@ -512,10 +552,25 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
     PROJECT: 'project',
+    PROPOSAL: 'proposal',
     REQUEST: 'request',
     RESOURCE: 'resource',
     THREAD: 'thread',
-    WELCOME: 'welcome',
+    WELCOME: 'welcome'
+  },
+
+  Proposal_Status: {
```
tibetsprague, line 562: oh hmm, I guess Type is not capitalized 🤷‍♂️ I guess leave it this way then

migrations/20240218134105_proposals.js
```diff
@@ -0,0 +1,47 @@
...
+  await knex.schema.table('posts', table => {
+    table.bigInteger('quorum')
+    table.text('proposal_status').index()
+    table.text('proposal_outcome').index()
+    table.text('proposal_type')
+    /*
+      added proposal_ prefix to status and outcome because they are very generic column names
+      and I don't want to confuse future developers as to why they are largely null
+    */
+    table.text('anonymity')
```
tibetsprague, line 11: wouldnt this be boolean?

tibetsprague (changes requested):
noice

api/graphql/index.js
```diff
@@ -456,9 +463,13 @@ export function makeMutations (expressContext, userId, isAdmin, fetchOne) {
     respondToEvent: (root, { id, response }) =>
       respondToEvent(userId, id, response),
 
+    setProposalOptions: (root, { postId, options }) => setProposalOptions({ userId, postId, options }),
+
     subscribe: (root, { groupId, topicId, isSubscribing }) =>
       subscribe(userId, topicId, groupId, isSubscribing),
 
+    swapProposalVote: (root, { postId, removeOptionId, addOptionId }) => swapProposalVote({ userId, postId, removeOptionId, addOptionId }),
```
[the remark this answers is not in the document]
thomasgwatson: Yeah its just a convenience mutation, to avoid having to handle making double calls on the front-end. And yes, only applicable for the single-vote situs

api/models/Post.js
thomasgwatson: Yeah, thats what I was going off

api/models/Post.js
thomasgwatson: Are there any that stand out to you as still possibly needed? I don't think so but wanted another pair of eyes on them

api/graphql/index.js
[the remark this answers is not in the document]
tibetsprague: but you wouldnt need a double call on the front end of addVote just removed the old one in the case of proposals that are set to only allow for one vote

api/models/Post.js
tibetsprague: hmm, no im not sure. probably good to get rid of? the socket one might need proposal stuff added to it but would have to be tested

migrations/20240218134105_proposals.js
thomasgwatson: I was leaving space for general anonymity:
- anonymous posts (and thus anonymous proposals)
- anonymous votes
- both

api/graphql/index.js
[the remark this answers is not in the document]
thomasgwatson: That would just keep adding votes to that proposal. Now we have to filter and sort them somewhere to determine what vote is someone's latest and we have to do that for every vote to figure out an accurate vote count. Seems like it causes a lot more hassle than it prevents

tibetsprague:
I don’t understand that. I thought we were going to have a setting that says whether a proposal can be voted on only once, or have multiple options voted on. In the case of the former then add vote could just delete the first vote and add the new one. In the case of the latter then it could add another vote. On Feb 27, 2024, at 6:15 PM, Tom Watson ***@***.***> wrote:﻿
@thomasgwatson commented on this pull request.



In api/graphql/index.js:
>      subscribe: (root, { groupId, topicId, isSubscribing }) =>
       subscribe(userId, topicId, groupId, isSubscribing),
 
+    swapProposalVote: (root, { postId, removeOptionId, addOptionId }) => swapProposalVote({ userId, postId, removeOptionId, addOptionId }),

That would just keep adding votes to that proposal. Now we have to filter and sort them somewhere to determine what vote is someone's latest and we have to do that for every vote to figure out an accurate vote count. Seems like it causes a lot more hassle than it prevents

thomasgwatson:
> I don’t understand that. I thought we were going to have a setting that says whether a proposal can be voted on only once, or have multiple options voted on. In the case of the former then add vote could just delete the first vote and add the new one. In the case of the latter then it could add another vote.

Yes, there will still be proposalType on the post object. I don't think having addProposalVote change behaviour is any better or worse than having swapVote as a convenience function. We need to collectively decide what proposalTypes there are (we have not defined them in the ticket, just discussed some different options. And the backend ideally should enforce the "only one vote" if that is the type they are.

thomasgwatson:
@tibetsprague what still needs to be addressed to push the backend for Proposals forward?
At the moment these changes offer:
- polls with unlimited options, and a single vote
- polls with unlimited options, and unlimited votes

So currently not covering vote limits beyond a single vote.

Also, anonymity as a string vs bool, to allow different types (anonymous voting, anonymous proposals, both anonymous would be the initial set). Also, I would suggest that if voting is anonymous, the results shouldn't show as the vote is in progress. And that the display of votes in progress is probably yet another separate field in the future.

api/graphql/index.js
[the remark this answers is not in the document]
tibetsprague: why would it keep adding votes, you first remove the old vote, then add the new one. but im not going to push it, your way is ok too

migrations/20240218134105_proposals.js
tibetsprague: i would make those separate fields if and when we need that. thats why i would call it anonymous_voting

tibetsprague:
i think this is good enough to move forward. i still think we should go with anonymous_voting and not mix that concept with a potential maybe future of wanting anonymous posts. as you build out the front-end I'm sure you will discover needed tweaks to the back-end so i would get moving on that

api/graphql/schema.graphql
```diff
@@ -21,7 +21,7 @@ type Query {
 
   # Query for PersonConnections
   connections(
-    # Number of PersonConnections to load
+    # Int of PersonConnections to load
```
tibetsprague, line 24: This comment seems more confusing. It is meant to say that this is the number of person connections to load. the type is already visible in the graphiql viewer so doesnt need to be repeated in the comment

api/graphql/schema.graphql
```diff
@@ -1642,8 +1641,12 @@ type Post {
   fulfilledAt: Date
   # Number of groups this has been posted to
   groupsTotal: Int
+  # Whether who votes on this proposal is visible or not
+  isAnonymousVote: Boolean
```
tibetsprague, line 1645: how is this different from anonymousVoting?

api/graphql/schema.graphql
```diff
@@ -1654,16 +1657,22 @@ type Post {
...
   myReactions: [Reaction]
-  # Whether the logged-in user has upvoted the post or not
-  myVote: Boolean
   # Number of people that have one or more reactions to a post
   peopleReactedTotal: Int
   # Number of total reactions on the post
   postReactionsTotal: Int
   # Number of total "members" of the post. Used for Projects only right now
   postMembershipsTotal: Int
+  # A proposal's outcome: 'cancelled', 'quorum-not-met', 'in-progress', 'successful', 'tie'
```
tibetsprague, line 1666: Looks like we never actually nailed down the possible outcomes. in the product spec it says it can set manually by the proposal creator, but it seems to me that a fixed set might be more useful? but we would definitely want a 'fail' option. Also what would a tie option mean? Seems like the outcome would either by a pass or a fail regardless of whether the number of votes in the options were tied

api/graphql/schema.graphql
```diff
@@ -1654,16 +1657,22 @@ type Post {
...
+  proposalOutcome: String
+  # The status of a proposal post: 'discussion', 'voting', 'completed'
+  proposalStatus: String
   # Link to a project management tool or service. Only used for Projects right now
   projectManagementLink: String
+  # The type of proposal: 'single', 'multiple-unrestricted', 'majority', 'consensus'
```
tibetsprague, line 1672: we havent discussed building in "types" of majority or consensus. And I think the mapping of number of votes to outcome (like majority) is not actually the same type of info as whether you get one or more votes. for now we are not building in voting details like majority of consensus, we are just letting the creator specify what the outcome will be based on the votes and then marking it as so. right?

api/graphql/schema.graphql
```diff
@@ -1985,6 +2020,8 @@ type Mutation {
   addPeopleToProjectRole(peopleIds: [ID], projectRoleId: ID): GenericResult
   # Add a Post to a Collection
   addPostToCollection(collectionId: ID, postId: ID): GenericResult
+  # Add a rvote to a proposal
```
tibetsprague, line 2023: extra r

api/graphql/schema.graphql
```diff
@@ -2206,6 +2249,8 @@ type Mutation {
   updateMembership(groupId: ID, slug: String, data: MembershipInput): Membership
   # Update a post you created
   updatePost(id: ID, data: PostInput): Post
+  # update options for a proposal
+  updateProposalOptions(postId: ID, options: [ProposalOptionInput]): [ProposalOption]
```
tibetsprague, line 2253: still not clear why you cant just use setProposalOptions again

api/models/Post.js
```diff
@@ -321,17 +323,80 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
     const updatedAttribs = pick(POSTS_USERS_ATTR_UPDATE_WHITELIST, omitBy(isUndefined, attrs))
-    return Promise.map(existingFollowers.models, postUser => postUser.updateAndSave(updatedAttribs, {transacting}))
+    return Promise.map(existingFollowers.models, postUser => postUser.updateAndSave(updatedAttribs, { transacting }))
+  },
+
+  async updateProposalOptions ({ options = [], userId, opts = { } }) {
+    const existingOptions = await this.proposalOptions().fetch({ transaction: opts.transacting, require: false })
+    const existingOptionIds = existingOptions.pluck('id')
+
+    // Delete ALL votes any time options are updated
```
tibetsprague, line 372: couldn't this all happen in setProposalOptions?

tibetsprague: I could see a case where folks want to add new options after voting has started. people do this for polls in FB all the time. Just an adding a new option wouldn't need to remove all old votes I dont think. but we should probably discuss whether we want this ability or not

api/models/Post.js
```diff
@@ -512,10 +577,34 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+
+  Proposal_Status: {
+    DISCUSSION: 'discussion',
+    COMPLETED: 'completed',
+    VOTING: 'voting',
+    CASUAL: 'casual'
+  },
+
+  Proposal_Outcome: {
+    CANCELLED: 'cancelled',
```
tibetsprague, line 595: what does canceled me?

api/models/Post.js
```diff
@@ -512,10 +577,34 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+    QUORUM_NOT_MET: 'quorum-not-met',
+    INCOMPLETE: 'incomplete',
+    IN_PROGRESS: 'in-progress',
+    SUCCESSFUL: 'successful',
+    TIE: 'tie'
```
tibetsprague, line 600: missing FAILED. dont think we need tie.

tibetsprague: or we just have a custom text field for outcome for now

api/models/Post.js
```diff
@@ -512,10 +577,34 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+  },
+
+  Proposal_Type: {
+    SINGLE: 'single',
+    MULTI_UNRESTRICTED: 'multi-unrestricted',
+    MAJORITY: 'majority', // one option must have more than 50% of votes for the proposal to 'pass',
```
tibetsprague, line 606: dont think we want/need these types

api/models/Post.js
```diff
@@ -597,6 +685,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
   }),
 
   create: function (attrs, opts) {
+    console.log(attrs,'attrs in create')
```
tibetsprague, line 688: console

api/models/post/setupPostAttrs.js
```diff
@@ -21,9 +23,27 @@ export default function setupPostAttrs (userId, params) {
...
   ))
 
-  return Promise.resolve(attrs)
+  let proposalAttrs = {}
+  let proposalStatus = params.startTime && new Date(Number(params.startTime)) < new Date() ? Post.Proposal_Status.VOTING : Post.Proposal_Status.DISCUSSION
+  if (params.endTime && new Date(Number(params.endTime)) < new Date()) proposalStatus = Post.Proposal_Status.COMPLETED
+  if (create) {
+    // if the startTime of the post is set and its before the current time in that timezone, then set the proposal status to VOTING
+    proposalAttrs = {
+      proposal_outcome: Post.Proposal_Outcome.INPROGRESS,
```
tibetsprague, line 37: should be IN_PROGRESS.
make sure this is tested

api/models/Post.js
```diff
@@ -163,6 +163,14 @@ module.exports = bookshelf.Model.extend(Object.assign({
     return this.hasMany(ProjectContribution)
   },
 
+  proposalOptions: function () {
+    return this.hasMany(ProposalOption)
```
tibetsprague, line 167: for performance purposes you can do     return this.get('type') === Post.Type.PROPOSAL ? this.hasMany(ProposalOption) : null
This means that if the front end loads proposalOptions for a stream of posts then it won't do the query to try and load them for non proposal posts, saving a ton of queries.

api/models/Post.js
```diff
@@ -163,6 +163,14 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+  },
+
+  proposalVotes: function () {
+    return this.hasMany(ProposalVote)
```
tibetsprague, line 171: for performance purposes you can do     return this.get('type') === Post.Type.PROPOSAL ? this.hasMany(ProposalVote) : null
This means that if the front end loads proposalOptions for a stream of posts then it won't do the query to try and load them for non proposal posts, saving a ton of queries.

tibetsprague (changes requested):
wahoo! still a couple things to sort out about proposal types and outcomes, but otherwise looking good.

api/graphql/schema.graphql
thomasgwatson: Currently, outcomes are not truly used anywhere; they are not displayed and the value being different doesn't change any behavior of the platform.

What did we settle on? The notes for the product meeting aren't clear. At the moment, once the voting ends, its basically up to the user to figure out the winner (which is fairly clear). We could refine results once we have a version that is out for general use, allowing the public to have a stronger sense of what would work for them

api/models/Post.js
```diff
@@ -164,11 +164,13 @@ module.exports = bookshelf.Model.extend(Object.assign({
   },
 
   proposalOptions: function () {
-    return this.hasMany(ProposalOption)
+    // return this.hasMany(ProposalOption)
```
tibetsprague, line 167: remove

api/models/Post.js
```diff
@@ -164,11 +164,13 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+    return this.get('type') === Post.Type.PROPOSAL ? this.hasMany(ProposalOption) : null
   },
 
   proposalVotes: function () {
-    return this.hasMany(ProposalVote)
+    // return this.hasMany(ProposalVote)
```
tibetsprague, line 172: remove

tibetsprague (approved):
i think this is looking good!

tibetsprague (approved):

thomasgwatson:
Not in production. Waiting on proposals in mobile to be rolled out. The
last time I checked, one could run the proposals mobile on the
non-proposals NODE, so the mobile one could be deployed first... but a
bunch of stuff has changed since then, so I wouldn't rely on that still
being true

lorenjohnson:
Excellent. Ok, was worried. So it sounds like a coordinated release may be the best call if we can swing it.

PR #1003, opened 2024-07-23
Node 1002 fix migrations and schema dump

### LINKED ISSUE #1002: Fix migrations and schema dump
- [ ] Change `SELECT pg_catalog.set_config('search_path', 'public', false);` back to `SELECT pg_catalog.set_config('search_path', '', false);` (removing the word public) 
- [ ] removing the line CREATE SCHEMA public;
- [ ] removing the whole CREATE PROCEDURE public.delete_user...

Some of these might be able to happen from flags on the `pg_dump` command, other things might have to happen using `sed` or something like that.

KevinTriplett:
Fixes #1002 

Two commits:
- fix the migration so that hopefully it won't break any dev machines current or future
- revise yarn script so it can dump a schema that runs under test and dev

I ran hylo-node tests and they worked after running `yarn dump`
I dropped my dev db and ran the READ.ME commands:
- createdb...
- cat schema...
- yarn knex seed:run
and I was able to run, create an account, creat a group, and create a proposal. Yippee!

### DIFF
--- a/migrations/20240218134105_proposals.js
+++ b/migrations/20240218134105_proposals.js
@@ -29,7 +29,7 @@ exports.up = async function (knex, Promise) {
     table.timestamp('created_at')
   })
 
-  await knex.raw('alter table proposal_options DROP CONSTRAINT proposal_options_post_id_foreign')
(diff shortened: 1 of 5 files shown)


thomasgwatson (approved):
lgtm

tibetsprague (approved):

PR #1004, opened 2024-07-23
Node 1001 Correct transacting arguments, tests passing

### LINKED ISSUE #1001: Correct the knex transacting arguments in post proposals
Per [this discussion on Hylo Dev Circle](https://www.hylo.com/groups/dev-circle/post/76194)

KevinTriplett:
Fixes #1001 

Existing pattern for Bookshelf arguments was inadvertently used when knex object is referenced.

Note:

The `opts.transacting ||= { transacting: false }` lines are a hack to tell knex when there are not transactions. This appears to only be the case in tests. This line is therefore necessary for tests that do not wrap database accesses in transactions. A proper fix would be to wrap tests in transactions, but that's more work than may be necessary.

Main disadvantage is reliance on `knex#transacting` not changing and breaking this. If that happens, it should only affect tests and not production.


thomasgwatson (approved):
lgtm! Thanks for getting to the bottom of this one

tibetsprague (approved):

PR #1005, opened 2024-07-29
Node 998 collective moderation

### LINKED ISSUE #998: Collective/community moderation for backend
All queries about posts are impacted:
- need to check if a post is flagged
- maybe just limit it to that for feed loads and rely on a click-through (and thus a load of PostDetail) to load all the extra stuff

posts table
- Need a short-hand way of knowing if a post is flagged. Do we do it in the db on the table and update constantly? Do we add it to the Post model as a getter and expose that way?
#### platform agreements
These now need to be tracked in the db. After thinking about it for a bit, it is pretty simple to just keep them as a separate table, instead of messing around with extending the agreements table. 

posts_users
- [ ] track whether a user has clicked-through on a post
- [ ] reset this value with each flag?
- [ ] build out mutations and model methods to allow this

How do we want to handle the database
- flag table (? a new table => moderation actions)
- activities table rehash?

I would also expect that moderation actions will be a record for other forms of moderation action in the future but we don't need to plan for that now

Need to capture
- relationships to agreements (many to many, so join table go whoosh. agreements_moderation_actions)
- text
- relationship to reporter (to one)
- relationship to post (to one)
- status (allow them to be cleared or reset)
- whether the reporter can be shown

Query changes (graphql)
- [ ] posts need to expose something concise (one field)
- [ ] posts need to expose a full relationship to moderation_actions
- [ ] ModerationActionQuerySet

Mutations
- [ ] allow moderator to clear flags
- [ ] allow creation of flag (no edits?)
- [ ] allow clearing of flag by reporter (no deletes?)
- [ ] record click-through by user

Notifications:
- [ ] Author and moderators are notified that a post has been flagged from their group
- [ ] Moderators get a notification if a post in their group has been flagged in another group (extra)

thomasgwatson:
Closes #998 

Going to start working on the EVO changes. Would anticipate that there will be further tweaks here but its substantive enough to open up commentary


api/models/ModerationAction.js
```diff
@@ -35,6 +40,63 @@ module.exports = bookshelf.Model.extend({
...
+  async sendEmailsForModerationAction ({ reporterId, postId, groupId, type = 'created' }) {
+    // type is 'created' or 'cleared'
+    // email reportee and reporter
+    const group = await Group.find(groupId)
+    const reporter = await User.find(reporterId)
+    const post = await Post.find(postId)
+    const reportee = post.relations.user
+    const link = await Frontend.Route.post(this.get('post_id'), group)
+    const reporterLocale = reporter.get('settings')?.locale || 'en'
+    const reporteeLocale = reportee.get('settings')?.locale || 'en'
```
tibetsprague, line 71: probably worth creating a User.getLocale() function

api/graphql/makeModels.js
```diff
@@ -293,6 +324,8 @@ export default function makeModels (userId, isAdmin, apiClient) {
         details: p => p.details(userId),
         isAnonymousVote: p => p.get('anonymous_voting') === 'true',
         myReactions: p => userId ? p.reactionsForUser(userId).fetch() : [],
+        // clickthrough: p => p.pivot && p.pivot.get('clickthrough'), // TODO COMOD: does not seem to work
```
tibetsprague, line 327: I do think this should work, I can try to play with it

api/graphql/schema.graphql
```diff
@@ -1563,6 +1579,43 @@ type MessageThreadQuerySet {
...
 }
 
+type ModerationAction {
+  id: ID
+  # The agreements that are linked to the moderation action
+  agreements: [Agreement]
+  # junk value, do not use
+  agreementsTotal: Int
+  # The person who took the action
+  reporter: Person
```
tibetsprague, line 1589: lets keep these alphabetical

api/graphql/schema.graphql
```diff
@@ -2882,6 +2957,16 @@ input MessageThreadInput {
   participantIds: [String]
 }
 
+# Data for a ModerationAction
+input ModerationActionInput {
+  postId: ID
```
tibetsprague, line 2962: alphabetical, and add comments

api/models/ModerationAction.js
```diff
@@ -0,0 +1,146 @@
...
+
+    const send = async (group, userIds, anonymous) => {
+      const text = await moderationAction.getMessageText({ group, groupId, anonymous })
+      return sendMessageFromAxolotl(userIds, text)
+    }
+    for (let group of groups) {
+      const moderators = await group.moderators().fetch()
+      await send(group, moderators.map(moderator => moderator.id), anonymous)
+    }
+    const shouldSendToAdmins = (process.env.HYLO_ADMINS &&
```
tibetsprague, line 136: didn't we decide to only send to group moderators for now, not to us?

api/services/Search/forModerationActions.js
```diff
@@ -0,0 +1,23 @@
...
+    qb.distinct()
+    qb.limit(opts.limit || 20)
+    qb.offset(opts.offset)
+    qb.orderBy('id', 'desc')
+
+    // this counts total rows matching the criteria, disregarding limit,
+    // which is useful for pagination
+    countTotal(qb, 'moderation_actions', opts.totalColumnName)
+
+    // TODO COMDO: ok, so I either need to make a direct relationship to groups or dig through agreements to find the group relationship
```
tibetsprague, line 14: direct relationship I say

test/setup/factories.js
```diff
@@ -5,6 +5,13 @@ import { ReadableStreamBuffer } from 'stream-buffers'
 import { faker } from '@faker-js/faker'
 
 module.exports = {
+  // agreement: attrs => {
```
tibetsprague, line 8: remove or use?

api/models/ModerationAction.js
thomasgwatson: So...

The current flagging setup actually sends us messages for certain report types. So I decided to replicate that here, since when we had the discussion it seems like we thought that we weren't actually getting messages from these flags

api/graphql/makeModels.js
thomasgwatson: sure thing

